PR #1, opened 2017-01-05
Adds basic workflow implementation

JosephSalisbury:
Towards unifying builds, my thinking is to have a simple binary that we ship in an image. This binary will largely replicate what we do in the `circle.yml` currently. We update the `circle.yml` to pull `latest` of the image, and then run it. This lets us update our build process in one place, and the builds will use it automatically.

This is a rough draft of the initial work for that binary. We define a `Workflow`, which is a series of `Tasks`. For example, `go test` and `go build` is one `Workflow`.
We also specify `Executors` to run these `Tasks`, a `DryRunExecutor` and `ShellExecutor` are provided.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,0 +1,34 @@
+package cmd
+
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/architect/workflow"
+)
+
+var (
+	buildCmd = &cobra.Command{
+		Use:   "build",
+		Short: "build the specified project",
+		Run:   runBuildCmd,
+	}
+
+	dryRun bool
+)
+
+func init() {
+	RootCmd.AddCommand(buildCmd)
+
+	buildCmd.Flags().BoolVar(&dryRun, "dry-run", false, "do not perform any actions")
+}
+
+func runBuildCmd(cmd *cobra.Command, args []string) {
+	executor := workflow.GetExecutor(dryRun)
+
+	build := workflow.NewGolangWorkflow()
+	if err := workflow.RunWorkflow(executor, build); err != nil {
+		log.Fatalln(err)
+	}
+}
--- a/workflow/executor.go
+++ b/workflow/executor.go
@@ -1,0 +1,52 @@
+package workflow
+
+import (
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Executor executes commands.
+type Executor interface {
+	// Run executes the given command, returning any errors.
+	RunCommand(arg ...string) error
+}
+
+// DryRunExecutor does nothing, successfully.
+type DryRunExecutor struct{}
+
+func (e *DryRunExecutor) RunCommand(arg ...string) error {
+	log.Printf("Executing: %v", arg)
+
+	return nil
+}
+
+// ShellExecutor executes commands in a shell.
+type ShellExecutor struct{}
+
+func (e *ShellExecutor) RunCommand(arg ...string) error {
+	command := strings.Join(arg, " ")
+	args := []string{"-c", command}
+
+	log.Printf("Executing: 'bash %v'", args)
+
+	cmd := exec.Command("bash", args...)
+
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+		return err
+	}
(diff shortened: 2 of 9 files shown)


cmd/build.go
```diff
@@ -0,0 +1,34 @@
...
+func init() {
+	RootCmd.AddCommand(buildCmd)
+
+	buildCmd.Flags().BoolVar(&dryRun, "dry-run", false, "do not perform any actions")
+}
+
+func runBuildCmd(cmd *cobra.Command, args []string) {
+	executor := workflow.GetExecutor(dryRun)
+
+	build := workflow.NewGolangWorkflow()
```
JosephSalisbury: In the future, we can infer the workflow from the project, or from configuration. Undecided. Currently aiming at golang projects only. I want to keep the scope small.

cmd/build.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"log"
+
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/architect/workflow"
+)
+
+var (
+	buildCmd = &cobra.Command{
```
JosephSalisbury: In the future, we'll probably have `deployCmd`

workflow/workflow.go
```diff
@@ -0,0 +1,26 @@
+package workflow
+
+type Workflow struct {
```
JosephSalisbury: I feel like this should be an interface, but a `Workflow` is basically just a list of `Tasks`, so it feels a little overkill. Thoughts?

workflow/workflow.go
```diff
@@ -0,0 +1,26 @@
...
+	Tasks []Task
+}
+
+func NewGolangWorkflow() Workflow {
```
JosephSalisbury: More advanced workflows will probably need to take some arguments for configuration.

workflow/executor.go
```diff
@@ -0,0 +1,52 @@
...
+	"log"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+// Executor executes commands.
+type Executor interface {
+	// Run executes the given command, returning any errors.
+	RunCommand(arg ...string) error
```
JosephSalisbury: While we only support shell commands now, I can see this extending in the future to include HTTP calls, and other I/O

workflow/gotest.go
```diff
@@ -0,0 +1,13 @@
+package workflow
+
+// GoTest runs `go test`.
+type GoTest struct {
+}
+
+func (t *GoTest) Run(e Executor) error {
+	if err := e.RunCommand("go", "test", "-v", "$(glide novendor)"); err != nil {
```
JosephSalisbury: I'm not certain we actually want shell expansion, it feels kind of weird - it forces us to always require a shell.

xh3b4sd:
#### 1. use open source
I am wondering why we should build the basics on our own. We should do some research on what is out there. I know there is grunt and gulp but this is nodejs. Even though I am kind of familiar with this eco system I fear its runtime deps. I found this so far. 

- https://github.com/dshills/goauto
- https://github.com/CenturyLinkLabs/golang-builder
- https://github.com/omeid/gonzo
- https://github.com/omeid/kargar

Especially kargar combined with gonzo could be interesting. Maybe we can make something up for our selves using these libraries and build it within the golang-builder image. Did not verify this could work. Just an idea to make my point.

---

#### 2. use our own tools
So yeah we need some CLI that wraps all the magic glue. I am wondering why not using the microkit basics? It provides the CLI which is extendable. In case we want to have feature x, y, z it will be there and we can use it for free. 

#### 3. mental model vs wording
This is rather minor but I was wondering about it. In the code there is a workflow and an executor defined. The workflow receives an executer and uses it to executes tasks. Since this is technically fine I would have expected that an executer executes a workflow. So it would be the other way around implementation wise. The executor should receive a workflow, shouldn't it? 

---

#### 4. mental model vs project structure
One last thing I had in mind while scanning the code. When we need to automate builds and when we have to define workflows for different languages and purposes, I would like to have this differentiation reflected by the package hierarchy. Something like this might make sense. 

```
.
└── architect
    └── workflow
        ├── go
        ├── python
        ├── shell
        └── ...
```

This is my feedback so far from the vacation brain. <3

JosephSalisbury:
This is over-engineering. Closing.

PR #3, opened 2017-01-31
Add basic binary that replicates circle.yml

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/980

Idea here is to have a binary that replicates what we do currently in the circle.yml. Deploying this to S3, and updating other circle.ymls will be in future PRs.

- [x] circle.yml
- [x] add kubectl image (see `giantswarm/kubectl:1.5.2`)
- [x] add support for deploying
- [x] refactor out docker run commands


JosephSalisbury:
```
2017/01/31 18:32:38 running command: docker-run
2017/01/31 18:32:38 docker run registry.giantswarm.io/giantswarm/kubernetesd:c8d7a5ec1d996076cdf63cbffcf2a0d471c1f51d --help
standard_init_linux.go:178: exec user process caused "no such file or directory"
2017/01/31 18:32:38 could not execute command 'docker-run': exit status 1
```
need to work out why this is happening

JosephSalisbury:
I think the image isn't build correctly - possibly we're in the wrong working directory

JosephSalisbury:
This problem seems to be confined to kubernetesd. Let's attack it when we get there.

circle.yml
```diff
@@ -0,0 +1,23 @@
...
+        - sudo apt-get update
+        - sudo apt-get install glide -y
+        - mkdir -p $HOME/go/src/github.com/$CIRCLE_PROJECT_USERNAME
+        - cp -rf $HOME/$CIRCLE_PROJECT_REPONAME $HOME/go/src/github.com/$CIRCLE_PROJECT_USERNAME
+
+test:
+    override:
+        - cd $GS_WD && go test -v $(glide novendor)
+        - cd $GS_WD && go build -a -v -tags netgo
+        - cd $GS_WD && ./architect --help
```
xh3b4sd: I cannot recall completely anymore. Why don't we build the docker image for architect here?

cmd/deploy.go
```diff
@@ -0,0 +1,137 @@
...
+
+	kubernetesResourcesDirectoryPath string
+	templatedResourcesDirectoryPath  string
+	removeResourceFilesAfterUse      bool
+)
+
+func init() {
+	RootCmd.AddCommand(deployCmd)
+
+	deployCmd.Flags().StringVar(&dockerEmail, "docker-email", "", "email to use to login to docker registry")
```
xh3b4sd: What do you think about the microkit style flag grouping? This would then be 

```
--service.docker.email
```

utils/glide.go
```diff
@@ -0,0 +1,47 @@
...
+			continue
+		}
+
+		files, err := ioutil.ReadDir(directory.Name())
+		if err != nil {
+			return nil, err
+		}
+
+		for _, file := range files {
+			if strings.HasSuffix(file.Name(), ".go") {
```
xh3b4sd: https://golang.org/pkg/path/filepath/#Ext is probably what we want to use.

utils/glide.go
```diff
@@ -0,0 +1,47 @@
...
+	for _, directory := range directories {
+		if !directory.IsDir() {
+			continue
+		}
+
+		if directory.Name() == "vendor" {
+			continue
+		}
+
+		files, err := ioutil.ReadDir(directory.Name())
```
xh3b4sd: Why are we doing this? We already iterate over all files/directories of the current working directory. Glide does not print nested dirs. What am I missing?

circle.yml
JosephSalisbury: I had issues getting a Docker image running other Docker images on CircleCI. So, I'm thinking we pull the binary down onto CircleCI, and not worry about an image at all.

cmd/deploy.go
JosephSalisbury: Personally, not _super_ fussed. Happy to change in the future. If you want it, open an issue for it, and I'll do it eventually <3

utils/glide.go
JosephSalisbury: We only care about directories that have go code inside of them. So, we look at all directories in the current working directory, and see which ones have go code inside them.

Random side though, the testing here is non-existent. I'll add an issue.

utils/glide.go
JosephSalisbury: Good tip. I'll update it.

circle.yml
xh3b4sd: With you.

utils/glide.go
JosephSalisbury: See https://github.com/giantswarm/architect/issues/4, https://github.com/giantswarm/architect/issues/5, and https://github.com/giantswarm/architect/issues/6

cmd/deploy.go
JosephSalisbury: See https://github.com/giantswarm/architect/issues/7

PR #8, opened 2017-02-10
GitHub releases

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/980

on merge to master, we create a tag + github release, and upload the binary to the release. we can then download this in other builds.


github-release.sh
```diff
@@ -0,0 +1,55 @@
+SHA=$1
+PERSONAL_ACCESS_TOKEN=$2
```
JosephSalisbury: `$PERSONAL_ACCESS_TOKEN` is a github personal access token for the user performing the build (we're using @taylorbot for this)

JosephSalisbury:
There is a small amount of "i'm not entirely certain if this works correctly till we merge to master", so please bear with me

circle.yml
```diff
@@ -21,3 +21,10 @@ test:
         - cd $GS_WD && go test -v $(glide novendor)
         - cd $GS_WD && go build -a -v -tags netgo
         - cd $GS_WD && ./architect --help
+        - cd $GS_WD && ./github-release.sh $CIRCLE_SHA1 $PERSONAL_ACCESS_TOKEN
```
xh3b4sd: You want to remove this before merging right?

github-release.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# fetch the release id for the upload
+RELEASE_ID=$(echo $release_output | jq '.id')
+
+echo "Upload binary to GitHub Release"
+upload_output=$(curl \
+    --header "Authorization: token ${PERSONAL_ACCESS_TOKEN}" \
+    --header "Content-Type: application/octet-stream" \
+    --data-binary @architect \
+    https://uploads.github.com/repos/giantswarm/architect/releases/${RELEASE_ID}/assets?name=architect
```
xh3b4sd: I think we should use a tool that acts as a Github API client in the long run. When this works now, fine.

circle.yml
JosephSalisbury: Good catch. Removing.

github-release.sh
JosephSalisbury: Totally agree - I doubt this will be _the way_ forever.

PR #12, opened 2017-02-16
Adds better support for running docker commands in circle

JosephSalisbury:


commands/commands.go
```diff
@@ -7,11 +7,53 @@ import (
...
 	Name string
 	Args []string
 }
 
+func NewDockerCommand(name string, config DockerCommandConfig) Command {
+	args := []string{
+		"docker", "run",
+	}
+
+	if os.Getenv("CIRCLECI") == "true" {
```
JosephSalisbury: This is the magic. If we're on circle, we use `--rm=false` to deal better with their btrfs issues

commands/commands.go
```diff
@@ -7,11 +7,53 @@ import (
 	"strings"
 )
 
+type DockerCommandConfig struct {
+	Volumes          []string
+	Env              []string
+	WorkingDirectory string
+	Image            string
+	Args             []string
```
xh3b4sd: It would make me so happy if this would be ordered based on some scheme. Like, the alphabet.

commands/commands.go
xh3b4sd: This sounds like we should write a comment down to make everybody who is reading the code aware of this reasoning.

commands/commands.go
JosephSalisbury: I've ordered them based on the order i was doing the arguments to the docker commands previously - i dislike alphabetising when there's already an order in my head

JosephSalisbury: And this matches the previous order we had in the makefiles :D

PR #14, opened 2017-02-17
Adds support for fetching kubernetes certificates from environment

JosephSalisbury:
When running in circle, we currently provide kubernetes
certificates via env vars.
This changeset lets us automatically fetch them,
which makes the circle.yml simpler

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -70,6 +70,15 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		log.Fatalf("could not template kubernetes resources: %v\n", err)
 	}
 
+	// When running in CircleCI, we can attempt to grab kubernetes certificates from the environment
+	if kubernetesCaPath == "" && kubernetesCrtPath == "" && kubernetesKeyPath == "" && os.Getenv("CIRCLECI") == "true" {
+		var err error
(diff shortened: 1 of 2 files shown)


cmd/deploy.go
```diff
@@ -70,6 +70,15 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		log.Fatalf("could not template kubernetes resources: %v\n", err)
 	}
 
+	// When running in CircleCI, we can attempt to grab kubernetes certificates from the environment
+	if kubernetesCaPath == "" && kubernetesCrtPath == "" && kubernetesKeyPath == "" && os.Getenv("CIRCLECI") == "true" {
```
xh3b4sd: IMO the validation should be that if one certificate is provided, there have to be all certificates provided. In case of circle it would be an issue to have only one given.

cmd/deploy.go
```diff
@@ -70,6 +70,15 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		log.Fatalf("could not template kubernetes resources: %v\n", err)
 	}
 
+	// When running in CircleCI, we can attempt to grab kubernetes certificates from the environment
```
xh3b4sd: That is a sentence and sentences end with a full stop. <3

PR #15, opened 2017-02-17
Changes build to also run version command

JosephSalisbury:


cmd/build.go
```diff
@@ -87,8 +87,16 @@ func runBuild(cmd *cobra.Command, args []string) {
 		},
 	}
 
-	dockerRun := commands.NewDockerCommand(
-		"docker-run",
+	dockerRunVersion := commands.NewDockerCommand(
+		"docker-run-version",
+		commands.DockerCommandConfig{
+			Image: fmt.Sprintf("%v/%v/%v:%v", registry, organisation, project, sha),
+			Args:  []string{"version"},
```
xh3b4sd: This is quite specific. We cannot expect everything built on circle providing a version command. Like this thing: https://github.com/giantswarm/k8s-setup-network-environment. It did not even had a help flag. WDYT? This can become quite messy.

JosephSalisbury: Adding the version command came from a previous conversation we had about running both `version` and `--help`, but this is probably only true for microkit stuff.

JosephSalisbury: Not certain yet. Let's get the G8S golang services sorted first, and then move forwards

Issue #16, opened 2017-02-20
Redact passwords in architect output


JosephSalisbury:
Architect prints command line args out - if we have a password in there, it'll be visible.

Not an issue for private repos, but is an issue for public ones

JosephSalisbury:
Note - renew docker passsword after this is done

PR #21, opened 2017-03-04
Redacts passwords in output

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/16

This changeset adds support for detecting passwords in
task name output, and automatically redacting them.

e.g;
```
2017/03/04 16:12:41 running command: docker-login: 'docker login --email=circleci@giantswarm.io --username=circleci --password=[REDACTED] registry.giantswarm.io'
```

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -38,8 +38,10 @@ func (c Command) String() string {
 
 		if requiresRedaction {
 			parts := strings.Split(arg, "=")
-			parts[1] = "[REDACTED]"
-			arg = parts[0] + "=" + parts[1]
+			if len(parts) == 2 {
+				parts[1] = "[REDACTED]"
+				arg = parts[0] + "=" + parts[1]
+			}
 		}
 
 		redactedArgs = append(redactedArgs, arg)
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


commands/commands.go
```diff
@@ -20,6 +21,33 @@ type Command struct {
...
+		"password",
+	}
+
+	redactedArgs := []string{}
+	for _, arg := range c.Args {
+		requiresRedaction := false
+
+		for _, redactionPhrase := range redactionPhrases {
+			if strings.Contains(arg, redactionPhrase) {
+				requiresRedaction = true
```
xh3b4sd: I don't see why this boolean variable is necessary. IMO you can simply do the magic in here and break the inner for loop when you are done.

commands/commands.go
```diff
@@ -20,6 +21,33 @@ type Command struct {
...
+			}
+		}
+
+		if requiresRedaction {
+			parts := strings.Split(arg, "=")
+			parts[1] = "[REDACTED]"
+			arg = parts[0] + "=" + parts[1]
```
xh3b4sd: Who guarantees that there are only arguments which are splittable by `=`, what then also results in at least two items? I can see this easily panicing. What happens if there would be a boolean password flag like `--password`?

commands/commands_test.go
```diff
@@ -1,11 +1,54 @@
...
+			command: Command{
+				Name: "docker-login",
+				Args: []string{"docker", "login", "--email=foo", "--password=bar"},
+			},
+			expectedString: "docker-login: 'docker login --email=foo --password=[REDACTED]'",
+		},
+		{
+			command: Command{
+				Name: "many-pass",
+				Args: []string{"foo", "--first-password=bar", "--second-password=baz"},
```
xh3b4sd: When we write tests for such cases we should try to cover as much as possible. There can also be boolean password flags and other word combinations like `--password` and `--password-first`. Does it also matter that there can be flags with only having one dash given like `-password`?

JosephSalisbury:
@xh3b4sd mind having another look? I fixed the panic

Issue #23, opened 2017-03-05
Email: EOF on docker login, during architect deploy


oponder:
The build failed while trying to deploy `cluster-service` during the `architect deploy` step:

```
./architect deploy
2017/03/05 14:11:07 running command: docker-login
2017/03/05 14:11:07 docker login --username=circleci --password=<REDACTED> registry.giantswarm.io
Email: EOF
2017/03/05 14:11:07 could not execute command 'docker-login': exit status 1

./architect deploy returned exit code 1

Action failed: ./architect deploy
```

I built it again with SSH to see what's going on, looks like it is waiting for user to input their email after the `docker login` command.

Looks like it is caused by this: https://github.com/giantswarm/architect/pull/22

oponder:
@JosephSalisbury I think we'll still need that email flag, despite it being deprecated?

Looks like circle also doesn't run the latest version of docker:

```
ubuntu@box2760:~/cluster-service$ docker -v
Docker version 1.9.1-circleci-cp-workaround, build 517b158
```

JosephSalisbury:
https://circleci.com/gh/giantswarm/cluster-service/193

I've fixed the `cluster-service` build. Closing <3

PR #25, opened 2017-03-05
Revert "Removes docker email from docker login"

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/23
Reverts giantswarm/architect#22


JosephSalisbury:
@xh3b4sd FYI

JosephSalisbury:
Basically, the docker version on circle is older than my local, and they still require the email flag. Pushing this forward so we don't block the builds.

Issue #24, opened 2017-03-05
README and project description could use some love


oponder:
We could start at least with a good description of what this is for. What tools it interfaces with (circleci), what problems it tries to solve... things like that.

https://thejunkland.com/blog/how-to-write-good-readme.html

PR #26, opened 2017-03-06
Improves readme a little

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/24

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,15 @@
 
 `architect` is a tool for managing Giant Swarm release engineering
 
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
yo @oponder where can i improve this further

oponder:
No need to add more for now, but its just nice to have a bit of an overview when you arrive at a project. I think a lot of our `golang` repos suffer from this.

JosephSalisbury:
if you find other situations where we don't write enough, open an issue <3

PR #27, opened 2017-03-07
Refactor to add workflow package

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1090

This PR introduces the `workflow` package, which determines which commands to run based on the state of the project directory.

E.g: if `main.go` is present, runs `go test` and `go build` in the build workflow, if `kubernetes/` directory is present, runs `kubectl apply` during deploy workflow

We also introduce [afero](https://github.com/spf13/afero) as a dependency, to ease testing workflows - this also allows us to more easily add tests for other packages in the future (particularly novendor, and kubernetes templating)

Note: because the package is named `workflow`, the PR is awkwardly split due to the massive `vendor` package :(

Next, I will test out support for deploying third party applications (nested kubernetes directories are not supported yet, for example)


JosephSalisbury:
Having issues with third party application projects with nested kubernetes directories. Will have a further look tomorrow, and clean up a little as well.

JosephSalisbury:
this pr definitely got away from me :D

cmd/deploy.go
```diff
@@ -66,93 +68,62 @@ func init() {
...
-	if err := utils.TemplateKubernetesResources(kubernetesResourcesDirectoryPath, templatedResourcesDirectoryPath, sha); err != nil {
-		log.Fatalf("could not template kubernetes resources: %v\n", err)
-	}
+	fs := afero.NewOsFs()
 
 	// When running in CircleCI, we can attempt to grab kubernetes certificates from the environment
 	if kubernetesCaPath == "" && kubernetesCrtPath == "" && kubernetesKeyPath == "" && os.Getenv("CIRCLECI") == "true" {
 		var err error
-		kubernetesCaPath, kubernetesCrtPath, kubernetesKeyPath, err = utils.FetchKubernetesCertsFromEnvironment(workingDirectory)
+		kubernetesCaPath, kubernetesCrtPath, kubernetesKeyPath, err = utils.FetchKubernetesCertsFromEnvironment(fs, workingDirectory)
```
xh3b4sd: `FetchKubernetesCertsFromEnvironment` -> `K8SCertsFromEnv`?

workflow/workflow.go
```diff
@@ -0,0 +1,311 @@
...
+	KubectlVersion                            string
+	KubernetesTemplatedResourcesDirectoryPath string
+
+	Goos          string
+	Goarch        string
+	GolangImage   string
+	GolangVersion string
+}
+
+func GetBuildWorkflow(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
```
xh3b4sd: `GetBuildWorkflow` -> `NewBuild`? Then it would read like `workflow.NewBuild`, which makes totally sense to me.

workflow/workflow.go
```diff
@@ -0,0 +1,311 @@
...
+				fmt.Sprintf("--email=%v", projectInfo.DockerEmail),
+				fmt.Sprintf("--username=%v", projectInfo.DockerUsername),
+				fmt.Sprintf("--password=%v", projectInfo.DockerPassword),
+				projectInfo.Registry,
+			},
+		}
+		w = append(w, dockerLogin)
+
+		dockerPush := commands.Command{
+			Name: "docker-push",
```
xh3b4sd: Having constants and a decent godoc for these names would make me happy. Looking at the tests I was wondering where they are coming from and I had to check the whole code.

cmd/build.go
```diff
@@ -34,80 +35,33 @@ func init() {
...
+		Organisation:     organisation,
+		Project:          project,
+		Sha:              sha,
+
+		Registry: registry,
+
+		Goos:          goos,
+		Goarch:        goarch,
+		GolangImage:   golangImage,
+		GolangVersion: golangVersion,
```
xh3b4sd: Only a general note to enrich the PR. I missed where all the variables are coming from based on this diff. I looked it up and these variables are actually assigned to flags of commands within the command package.

workflow/workflow.go
JosephSalisbury: Yeah, this is a good idea. Will update.

workflow/workflow.go
JosephSalisbury: Yep, makes sense. Will update.

cmd/deploy.go
JosephSalisbury: Yeah, that's way nicer. Will update.

PR #30, opened 2017-03-08
Adds support for templating nested kubernetes directories

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1090

This PR adds support for templating singly nested kubernetes directories - for example:
```
kubernetes/
   node_exporter/
      deployment.yml
   prometheus/
      deployment.yml
```

We don't currently need to replace docker tags for third party applications, but the templating happens anyway.


utils/kubernetes.go
```diff
@@ -10,27 +10,79 @@ import (
...
-			return err
-		}
+func TemplateFile(fs afero.Fs, path, templatedPath, sha string) error {
+	contents, err := afero.ReadFile(fs, path)
+	if err != nil {
+		return err
 	}
 
-	files, err := afero.ReadDir(fs, kubernetesResourcesDirectoryPath)
+	templatedContents := strings.Replace(string(contents), "%%DOCKER_TAG%%", sha, -1)
```
xh3b4sd: Do we want to use the proper golang template language? Lets create an issue if you agree.

utils/kubernetes.go
```diff
@@ -10,27 +10,79 @@ import (
...
+					})
+				}
+			}
+		} else {
+			filesToTemplate = append(filesToTemplate, FileToTemplate{
+				path:          filepath.Join(resourcesDir, fileInfo.Name()),
+				templatedPath: filepath.Join(templatesDir, fileInfo.Name()),
+			})
+		}
+	}
```
xh3b4sd: We want to use https://golang.org/pkg/path/filepath/#Walk.

utils/kubernetes.go
```diff
@@ -10,27 +10,79 @@ import (
...
+
+	// Create the templated resources directory, if it does not exist
+	if _, err := fs.Stat(templatesDir); os.IsNotExist(err) {
+		if err := fs.Mkdir(templatesDir, 0644); err != nil {
 			return err
 		}
+	}
```
xh3b4sd: Why are we creating a directory? We only have to overwrite the existing files and eventually checkout all changes after the build.

utils/kubernetes_test.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+
+				if err := afero.WriteFile(
+					fs,
+					filepath.Join(resourcesPath, "foo/", "deployment.yml"),
+					testData,
+					filePermission,
+				); err != nil {
+					return err
+				}
```
xh3b4sd: FYI I do not like this style.

JosephSalisbury:
Yeah, I think I overthought this. My original thinking was to create the secondary directory (`kubernetes-templated/`, so we weren't overwriting the `kubernetes` directory, but that really doesn't matter on circle, because we throw everything away afterwards anyway.

I'll remove the secondary directory, it should simplify stuff a lot

utils/kubernetes.go
JosephSalisbury: I guess? Not certain. I'll open an issue. We need this straight in the future, for certain. See https://github.com/giantswarm/giantswarm/issues/1306 for tracking

JosephSalisbury:
@xh3b4sd mind having another look? `Walk` was a good shout, and I removed the secondary directory, it cleaned up a lot

PR #31, opened 2017-03-08
Improves workflow printing

JosephSalisbury:
This makes me happy

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -29,12 +29,16 @@ var (
 type Workflow []commands.Command
 
 func (w Workflow) String() string {
-	cmdStrings := []string{}
-	for _, cmd := range w {
-		cmdStrings = append(cmdStrings, "\t"+cmd.String())
+	if len(w) == 0 {
+		return "{}"
 	}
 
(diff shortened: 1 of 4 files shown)


workflow/workflow.go
```diff
@@ -29,12 +29,16 @@ var (
...
+	}
+
 	cmdStrings := []string{}
 	for _, cmd := range w {
-		cmdStrings = append(cmdStrings, "\t"+cmd.String())
+		cmdStrings = append(cmdStrings, "\t"+cmd.String()+"\n")
 	}
 
-	return "{\n" + strings.Join(cmdStrings, "\n") + "\n}"
+	return fmt.Sprintf("{\n%v}", strings.Join(cmdStrings, ""))
```
xh3b4sd: Removing the newline from line 38 and at it to here to the format string would make me happy. Then it is this. 

```
"{\n%v\n}"
````

JosephSalisbury: Yeah, but then we end up with everything being one big line :D

JosephSalisbury: Consider multiple commands in a workflow

PR #32, opened 2017-03-09
Adds support to workflow package to deploy to multiple clusters

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1300

This provides no new functionality, but prepares the `workflow` package for deploying to multiple clusters (with clusters being an array, instead of a set of strings)

Splitting it up into 2 PRs because this one is already pretty chunky, and I can see wiring the frontend being messy too - I'll do that next.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -96,12 +96,16 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		DockerUsername: dockerUsername,
 		DockerPassword: dockerPassword,
 
-		KubernetesApiServer:                       kubernetesApiServer,
-		KubernetesCaPath:                          kubernetesCaPath,
-		KubernetesCrtPath:                         kubernetesCrtPath,
-		KubernetesKeyPath:                         kubernetesKeyPath,
-		KubectlVersion:                            kubectlVersion,
 		KubernetesTemplatedResourcesDirectoryPath: templatedResourcesDirectoryAbsolutePath,
+		KubernetesClusters: []workflow.KubernetesCluster{
+			workflow.KubernetesCluster{
+				ApiServer:      kubernetesApiServer,
+				CaPath:         kubernetesCaPath,
+				CrtPath:        kubernetesCrtPath,
+				KeyPath:        kubernetesKeyPath,
+				KubectlVersion: kubectlVersion,
+			},
+		},
 	}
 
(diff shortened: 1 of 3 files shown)


workflow/workflow.go
```diff
@@ -223,6 +227,12 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		return nil, err
 	}
 	if dockerFileExists {
+		if projectInfo.Sha == "" {
```
JosephSalisbury: We were missing the check for the sha here, noticed it while updating the deploy tests with a project info per test. Registry moved as well for consistency with project info flags.

cmd/deploy.go
```diff
@@ -96,12 +96,16 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		DockerUsername: dockerUsername,
 		DockerPassword: dockerPassword,
 
-		KubernetesApiServer:                       kubernetesApiServer,
-		KubernetesCaPath:                          kubernetesCaPath,
-		KubernetesCrtPath:                         kubernetesCrtPath,
-		KubernetesKeyPath:                         kubernetesKeyPath,
-		KubectlVersion:                            kubectlVersion,
 		KubernetesTemplatedResourcesDirectoryPath: templatedResourcesDirectoryAbsolutePath,
+		KubernetesClusters: []workflow.KubernetesCluster{
```
JosephSalisbury: This gives the same result right now, but allows me to add support for more clusters next

xh3b4sd: Where do these clusters come from? Right now one is hard coded. That should be configurable.

workflow/workflow.go
xh3b4sd: For my interest, why was the sha missing? When can this happen?

cmd/deploy.go
JosephSalisbury: > Splitting it up into 2 PRs because this one is already pretty chunky, and I can see wiring the frontend being messy too - I'll do that next.

JosephSalisbury: So, it will be configurable, in the next PR

workflow/workflow.go
JosephSalisbury: When I do coding bad :(

Issue #4, opened 2017-02-01
Add tests for glide novendor


JosephSalisbury:
The glide novendor function needs tests

PR #35, opened 2017-03-10
Adds tests, refactors novendor

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/4

I found a subtle bug where a package containing no golang files, but containing a directory with golang files in, was not correctly listed as a package to test. This is fixed now.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -100,7 +100,7 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 			return nil, fmt.Errorf("golang version cannot be empty")
 		}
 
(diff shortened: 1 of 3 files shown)


workflow/workflow.go
```diff
@@ -100,7 +100,7 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 			return nil, fmt.Errorf("golang version cannot be empty")
 		}
 
-		testPackageArguments, err := utils.NoVendor(projectInfo.WorkingDirectory, fs)
+		testPackageArguments, err := utils.NoVendor(fs, projectInfo.WorkingDirectory)
```
JosephSalisbury: This is me satisfying my inner monk - all other functions take afero as their first argument, so i made the signature of `NoVendor` consistent too

PR #36, opened 2017-03-17
workflow: support dependencies and add go-generate command

iaguis:
Some projects (like the aws-operator) use go generate at build-time.

Adding a step to run "go generate".

Also, adding a "go get" step to get build-time dependencies. These can
be specified in a CLI argument.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,15 +6,15 @@ machine:
         GOOS: "linux"
         GOARCH: "amd64"
         GOPATH: "$HOME/go"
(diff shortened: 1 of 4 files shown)


iaguis:
I'm not sure this is an approach you're ok with, please let me know what you think.

JosephSalisbury:
I don't want to use gobindata. That feels super weird. I saw some conversation about removing our usage of removing it. That makes way more sense to me.

Let's talk about why we're doing this.

PR #37, opened 2017-03-20
Adds support for ignoring directories starting with underscore

JosephSalisbury:
Golang considers directories that begin with an underscore to
not be packages. This behaviour is also present in glide.
This changeset brings the behaviour of architect up to match
glide, by ignoring directories starting with underscores.

See: https://golang.org/pkg/go/build/#Context.Import


JosephSalisbury:
Towards fixing https://circleci.com/gh/giantswarm/testbot/54?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #42, opened 2017-03-23
Configuration

JosephSalisbury:
Towards overall system configuration. This is currently needed by architect to handle configuration differences between leaseweb and AWS.

The structure and data are currently in `architect`. The structure will most likely be moved into a separate library in the future, and `deployment-agent` will hold the data, via configmap. In the future, we would like to use `configd` to hold the configuration data.

This PR presents a draft of the configuration structure, as well as an example of how the leaseweb installation would be configured.

I would also recommend having a look with godoc, I have tried to make that as useful as possible.


configuration/configuration.go
```diff
@@ -0,0 +1,25 @@
...
+type Installation struct {
+	V1
+}
+
+// V1 is the version 1 of the configuration structure.
+type V1 struct {
+	apiservices.APIServices
+	guestclusters.GuestClusters
+	monitoring.Monitoring
+	vault.Vault
```
JosephSalisbury: I don't like vault being kind of rogue here. Thoughts on a group for it would be appreciated.

configuration/guestclusters/guestclusters.go
```diff
@@ -0,0 +1,17 @@
+// Package guestclusters provides configuration structures for guest clusters.
+package guestclusters
+
+// GuestClusters holds the configuration for the installation's
+// guest clusters.
+type GuestClusters struct {
```
JosephSalisbury: I feel like this needs breaking down, but can't see it yet. Alternatively, we add KVM / AWS specific structures / fields later, and this is already the top level object.

installation/leaseweb.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/giantswarm/architect/configuration/guestclusters"
+	"github.com/giantswarm/architect/configuration/monitoring"
+	"github.com/giantswarm/architect/configuration/vault"
+)
+
+var Leaseweb = configuration.Installation{
+	V1: configuration.V1{
+		APIServices: apiservices.APIServices{
+			GSAPI: apiservices.GSAPI{
+				Address: url.URL{
```
JosephSalisbury: I appreciate that `url.URL` is more verbose than `string`, but I prefer it. Future configuration systems (e.g: `deployment-agent` / `configd`) coudl use `string`, and handle the potential parse error.

JosephSalisbury:
Going for a big team review, I want people's eyes and opinions. We need to make sure that the start here is well designed, to avoid configuration complexity issues in the future.

JosephSalisbury:
Don't feel like that means I *need* everyone to review, just making it visible :D

configuration/configuration.go
rossf7: What about authentication or auth to keep it shorter? In the future could integrations like LDAP go in there?

configuration/vault/vault.go
```diff
@@ -0,0 +1,21 @@
...
+
+// Vault holds the configuration for the installation's Vault setup
+type Vault struct {
+	// Address is the URL of the Vault API.
+	// e.g: 'https://leaseweb-vault-private.giantswarm.io:8200'
+	Address url.URL
+
+	// CaTTL is the TTL for CAs.
+	// e.g: `86400h`
+	CaTTL string
```
rossf7: On cert-operator @xh3b4sd wanted me to break these down further to Vault.CA.TTL etc. I have an issue to rename them :)

configuration/configuration.go
puja108: `auth` sounds good

configuration/apiservices/apiservices.go
```diff
@@ -0,0 +1,18 @@
+// Package apiservices provides configuration structures for API services.
+package apiservices
+
+import (
+	"net/url"
+)
+
+// APIServices holds configuration for API services.
+type APIServices struct {
```
xh3b4sd: Talking about structure, for me it would be more like this. 

```
configuration/service/api/gs
Configuration.Service.API.GS
```

I try to get the bigger picture. Then I wonder what `service` means. This term is super overloaded. Let me scribble a bit. I am sure we can easily overdo it and we simply have to commit to some structure. Anyway, here my scribble. 
```
configuration/cluster/guest
configuration/cluster/host
configuration/cluster/guest/ingress-controller
configuration/cluster/host/ingress-controller
configuration/giantswarm/api
configuration/giantswarm/kubernetesd
configuration/auth/vault
configuration/monitoring/prometheus
configuration/monitoring/testbot
```

---

When we structure the configuration and need golang structures I want to suggest to structure and separate the directories and flies accordingly. Check for instance [the directory/file structure of the flags in kubernetesd](https://github.com/giantswarm/kubernetesd/tree/3c6c8eb6b21c7c6f8841e561a9f7ca7797e88e31/flag). It is a little bit of work to set it up and it is a little bit verbose, but it is super comfortable to extend and work with. That way you will also not have weird configuration/variable names or inconsistencies because of weird collisions later on when it is structured in such a pedantic way.

JosephSalisbury:
I have moved Vault into an `auth` group, and am now using `time.Duration` instead of strings where it makes sense

JosephSalisbury:
I have also broken down the vault struct, and added even more comments

configuration/cluster/guest.go
```diff
@@ -0,0 +1,18 @@
+// Package cluster provides configuration structures for clusters.
+package cluster
+
+import (
+	"github.com/giantswarm/architect/configuration/cluster/hyperkube"
```
xh3b4sd: The hyperkube block is inside the guest block. The package path should reflect that to be consistent.

configuration/auth/vault/vault.go
```diff
@@ -0,0 +1,32 @@
...
+
+	// CA is the configuration for Vault CAs.
+	CA
+
+	// Token is the configuration for Vault tokens.
+	Token
+}
+
+// CA holds configuration for a certificate authority.
+type CA struct {
```
xh3b4sd: I think this should be in a separate package in case you want to follow your own rules.

PR #44, opened 2017-03-23
Adds support for resource templating

JosephSalisbury:
Adds support for templating using the new `configuration` package, and wires that into the current templating support.


JosephSalisbury:
See https://github.com/giantswarm/giantswarm/issues/1332 for tracking updating away from `%%DOCKER_TAG%%` - we keep backwards compatability with that for now

JosephSalisbury:
https://github.com/giantswarm/api/issues/339 and https://github.com/giantswarm/g8s-prometheus/issues/7 are updates for `api` and `g8s-prometheus` that update their templating with regards to ingress tags

template/utils.go
```diff
@@ -0,0 +1,26 @@
+package template
```
xh3b4sd: IMO `utils.go` implies that something is wrong with the mental model.

workflow/clusters.go
```diff
@@ -1,26 +1,44 @@
 package workflow
 
-import "github.com/spf13/afero"
+import (
+	"github.com/giantswarm/architect/configuration"
+	"github.com/giantswarm/architect/installation"
+	"github.com/spf13/afero"
+)
+
+type KubernetesCluster struct {
```
JosephSalisbury: This has been moved here from `workflow.go`, I forgot about it during the previous refactor

workflow/clusters.go
```diff
@@ -34,11 +52,13 @@ func ClustersFromEnv(fs afero.Fs, workingDirectory string) ([]KubernetesCluster,
 
 		newCluster := KubernetesCluster{
 			ApiServer:      configuredCluster.ApiServer,
-			IngressTag:     configuredCluster.IngressTag,
+			Prefix:         configuredCluster.EnvVarPrefix,
```
JosephSalisbury: We now use the env var prefix for the templated directory name

workflow/clusters_test.go
```diff
@@ -0,0 +1,200 @@
+package workflow
```
JosephSalisbury: `clusters_test.go` is me fixing the fact i forgot to move this as well

workflow/workflow_test.go
```diff
@@ -370,175 +366,3 @@ func TestGetDeployWorkflow(t *testing.T) {
 		}
 	}
 }
-
-func TestClustersFromEnv(t *testing.T) {
```
JosephSalisbury: Moved to `clusters_test.go`

PR #47, opened 2017-04-05
Removes unnecessary port from vault address

JosephSalisbury:
We're serving Vault behind an ELB, which handles on 443.


JosephSalisbury:
We'll need to redeploy anything which uses this value afterwards, which i think is just cluster-operator and cert-operator. i'll take care

PR #49, opened 2017-04-11
Add config for services that Happa relies on

oponder:
We'll need this config per installation. It does make me want to move stuff that passage does more into API, as well as Desmotes. or at least put Passage and Desmotes 'behind' API.

But for now we'd need it. @JosephSalisbury what do you think?

### DIFF
--- a/configuration/giantswarm/giantswarm.go
+++ b/configuration/giantswarm/giantswarm.go
@@ -2,9 +2,13 @@
 package giantswarm
 
 import "github.com/giantswarm/architect/configuration/giantswarm/api"
+import "github.com/giantswarm/architect/configuration/giantswarm/passage"
+import "github.com/giantswarm/architect/configuration/giantswarm/desmotes"
 
 // GiantSwarm holds configuration for GiantSwarm services.
 type GiantSwarm struct {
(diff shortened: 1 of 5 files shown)


configuration/giantswarm/giantswarm.go
```diff
@@ -2,9 +2,13 @@
...
 
 import "github.com/giantswarm/architect/configuration/giantswarm/api"
+import "github.com/giantswarm/architect/configuration/giantswarm/passage"
+import "github.com/giantswarm/architect/configuration/giantswarm/desmotes"
 
 // GiantSwarm holds configuration for GiantSwarm services.
 type GiantSwarm struct {
 	// API holds configuration for the GiantSwarm API.
 	api.API
+	passage.Passage
```
JosephSalisbury: Can we have a comment here similar to the API one

configuration/giantswarm/giantswarm.go
```diff
@@ -2,9 +2,13 @@
...
+	desmotes.Desmotes
```
JosephSalisbury: And a similar comment here

oponder:
rfr again.

PR #55, opened 2017-04-26
added virtualization provider

xh3b4sd:
This PR provides information of the currently used virtualization provider for each installation config. We need this to know which kind of validation and internal business logic to chose from within our G8S microservices.


configuration/provider.go
```diff
@@ -0,0 +1,8 @@
+package configuration
+
+type provider string
+
+const (
+	ProviderAWS provider = "aws"
```
JosephSalisbury: I would like docstring here

configuration/provider.go
```diff
@@ -0,0 +1,8 @@
...
+	ProviderKVM provider = "kvm"
```
JosephSalisbury: I would like docstring here

configuration/provider.go
```diff
@@ -0,0 +1,8 @@
+package configuration
```
JosephSalisbury: Need package docstring

xh3b4sd: No it doesn't. The package docs are in `configuration.go`. This is not a new package.

configuration/provider.go
xh3b4sd: Done.

configuration/provider.go
xh3b4sd: Done.

PR #56, opened 2017-04-27
added draft for instance type configuration

xh3b4sd:
Towards https://github.com/giantswarm/api/issues/350.

### DIFF
--- a/configuration/provider/aws/ec2/ec2.go
+++ b/configuration/provider/aws/ec2/ec2.go
@@ -1,0 +1,12 @@
+// Package ec2 provides configuration structures for the EC2 specific settings.
+package ec2
+
+import (
+	"github.com/giantswarm/architect/configuration/provider/aws/ec2/instance"
+)
+
+// EC2 holds configuration for the EC2 specific settings.
+type EC2 struct {
+	// Instance holds configuration for the allowed settings on EC2.
+	instance.Instance
+}
--- a/configuration/provider/aws/ec2/ec2.go
+++ b/configuration/provider/aws/ec2/ec2.go
@@ -7,6 +7,6 @@ import (
 
 // EC2 holds configuration for the EC2 specific settings.
 type EC2 struct {
-	// Instance holds configuration for the allowed settings on EC2.
+	// Instance holds configuration for the instance type settings on EC2.
 	instance.Instance
 }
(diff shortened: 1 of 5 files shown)


installation/aws.go
```diff
@@ -91,6 +94,16 @@ var AWS = configuration.Installation{
...
 
 		Provider: provider.Provider{
+			AWS: aws.AWS{
+				EC2: ec2.EC2{
+					Instance: instance.Instance{
+						Allowed: instance.Allowed(
+							instance.TypeM3Large,
+						),
+						Available: instance.Available(),
+					},
```
xh3b4sd: Note that the structure derives a bit as we discussed. This is due to the structure of the code. I want to have a private `kind` type for the instance type constants to ensure they are only used as we defined them in the `instance` package. This would not have been possible with the structure we came up with before, because an import circle would have been caused. I went with this and like it even more.

configuration/provider/aws/ec2/ec2.go
```diff
@@ -0,0 +1,12 @@
+// Package ec2 provides configuration structures for the EC2 specific settings.
+package ec2
+
+import (
+	"github.com/giantswarm/architect/configuration/provider/aws/ec2/instance"
+)
+
+// EC2 holds configuration for the EC2 specific settings.
+type EC2 struct {
+	// Instance holds configuration for the allowed settings on EC2.
```
rossf7: Nitpick but I'd change comment to say this is configuration for an EC2 instance type.

configuration/provider/aws/ec2/instance/instance.go
```diff
@@ -0,0 +1,196 @@
...
+package instance
+
+// kind is a private type to ensure only instance types defined in this package
+// can be applied to installation configurations. That prevents other packages
+// screwing around with instance type configurations.
+type kind string
+
+// The following list of instance types was fetched from the following reference.
+//
+//     curl -s https://raw.githubusercontent.com/powdahound/ec2instances.info/master/www/instances.json | jq '.[].instance_type'
```
rossf7: I like that we're using the github repo. Using curl and jq makes sense as these don't change often. We can always automate later and really AWS should provide an API for this.

configuration/provider/aws/ec2/ec2.go
xh3b4sd: Done.

installation/aws.go
rossf7: I like this structure and the Allowed and Available names.

PR #58, opened 2017-04-27
fixed guest cluster structure

xh3b4sd:
This is just for sanity. I noticed the directory structure does not follow the golang structs structure as the others do. This should not break anything.


configuration/guest/hyperkube/hyperkube.go
```diff
@@ -0,0 +1,17 @@
+// Package hyperkube provides configuration structures for Hyperkube.
+package hyperkube
+
+// kind is a private type to ensure only versions defined in this package can
+// be applied to installation configurations. That prevents other packages
+// screwing around with version configurations.
+type kind string
+
+const (
+	Version kind = "v1.5.2_coreos.2"
```
xh3b4sd: Note that `v1.5.2_coreos.0` was configured before. I found that we are actually using this updated version. See also https://github.com/giantswarm/kubernetesd/pull/118.

rossf7: I like that we're updating and hyperkube upgrades won't happen that often. But this feels a bit hidden to me. Could the constant be in installation?

xh3b4sd: I do not think it belongs to there. We should then move the provider kinds and instance types as well. This is the same concept, isn't it?

rossf7: OK that makes sense. I just wanted to check as I'm still getting my head around our infrastructure as code ;)

xh3b4sd: No worries buddy.

PR #60, opened 2017-04-28
removed unused cluster worker type from happa config

xh3b4sd:


configuration/giantswarm/happa/happa.go
```diff
@@ -10,8 +10,4 @@ type Happa struct {
 	// Address is the URL to Happa.
 	// e.g: 'https://happa-g8s.giantswarm.io'
 	Address url.URL
-
-	// CreateClusterWorkerType controls the type of form to show when creating a
-	// cluster. Valid values are 'aws' or 'kvm'.
-	CreateClusterWorkerType string
```
xh3b4sd: Happa uses the more general provider kind. This is not needed anymore.

PR #63, opened 2017-05-08
Adds support for pushing images to quay

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1087

We add support for specifying QUAY_USERNAME and QUAY_PASSWORD
env vars, which if specified, forces architect to use the quay
registry.
We can add these env vars as needed to each project one-by-one
as they move.

In the future, after everything has been moved to quay, we can
clean up these spare env vars, and have one registry again.


cmd/build.go
```diff
@@ -35,6 +36,10 @@ func init() {
 }
 
 func runBuild(cmd *cobra.Command, args []string) {
+	if os.Getenv("QUAY_USERNAME") != "" {
+		registry = "quay.io"
+	}
```
xh3b4sd: Isn't there `DOCKER_USERNAME` and `QUAY_USERNAME` at the same time then? How does that work with only having registry being a string? Don't we provide multiple registries? How should that work?

cmd/deploy.go
```diff
@@ -67,6 +73,10 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		log.Fatalf("could not get absolute path for resources directory: %v\n", err)
 	}
 
+	if os.Getenv("QUAY_USERNAME") != "" {
```
xh3b4sd: Saw that several times now. Would it make sense to have a more decent and reusable method for the quay check? We should then also be more strict and make sure that the env vars are there as we expect them. But I see that we might not want to overdo it now when we throw this logic away with pull based deployments. Uncertain.

cmd/build.go
JosephSalisbury: We don't provide multiple registries. When we want to move a project to quay, we add `QUAY_USERNAME` and `QUAY_PASSWORD` to circle, which forces architect to use the quay registry. Once all the projects have been moved, we'll drop the `DOCKER_*` env vars, and clean up the conditionals.

JosephSalisbury: This is largely to avoid breaking everything at once, there's a semi-gross transitional stage as we move things.

cmd/deploy.go
JosephSalisbury: We're throwing this away at the end of https://github.com/giantswarm/giantswarm/issues/1087, no strong feelings.

PR #78, opened 2017-05-26
Network host

oponder:
In order to do some integration testing, it would be handy if the architect build could reach services that are started up on the host. 

That way we don't have to try and do something more complex like passing in links as part of `architect build`


oponder:
Not sure why i called it "Net".
"Network" would be more in line with the actual flag.

oponder:
screen shot 2017-05-26 at 15 56 31

PR #85, opened 2017-05-29
Removes failing version check

JosephSalisbury:


JosephSalisbury:
We can do this in the same PR, actually

PR #88, opened 2017-05-30
add kubectl config

asymmetric:
With this PR, `architect` sets the version of the kubectl image.

I decided to create a separate package for `kubectl` config. Not sure if it makes more sense to have it within the `kubernetes` package.

Towards  https://github.com/giantswarm/k8scloudconfig/issues/136

### DIFF
--- a/configuration/guest/kubectl/kubectl.go
+++ b/configuration/guest/kubectl/kubectl.go
@@ -1,0 +1,17 @@
+// Package kubectl provides configuration structures for Kubectl.
+package kubectl
+
+// kind is a private type to ensure only versions defined in this package can
+// be applied to installation configurations. That prevents other packages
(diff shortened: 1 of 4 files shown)


configuration/guest/kubectl/kubectl.go
```diff
@@ -0,0 +1,17 @@
+// Package kubectl provides configuration structures for Kubectl.
+package kubectl
+
+// kind is a private type to ensure only versions defined in this package can
+// be applied to installation configurations. That prevents other packages
+// screwing around with version configurations.
+type kind string
+
+const (
+	Version kind = "1afa480ffb2912fe3605e84fd392c5bd1c9f48b9"
```
JosephSalisbury: A comment for the internal kubectl version would be nice

PR #91, opened 2017-05-31
Set certificate and token TTLs to 6 months

rossf7:
Towards giantswarm/giantswarm#1504

Until we have automated renewal of certs we want to make them valid for 6 months. This PR adds a Certificate block to Vault and sets both it and the token TTL to 6 months.

### DIFF
--- a/installation/aws.go
+++ b/installation/aws.go
@@ -36,8 +36,11 @@ var AWS = configuration.Installation{
 				CA: vault.CA{
 					TTL: 10 * 365 * 24 * time.Hour,
 				},
+				Certificate: vault.Certificate{
+					TTL: 26 * 7 * 24 * time.Hour,
+				},
 				Token: vault.Token{
-					TTL: 30 * 24 * time.Hour,
+					TTL: 26 * 7 * 24 * time.Hour,
 				},
 			},
 		},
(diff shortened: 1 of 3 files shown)


installation/aws.go
```diff
@@ -36,8 +36,11 @@ var AWS = configuration.Installation{
 				CA: vault.CA{
 					TTL: 10 * 365 * 24 * time.Hour,
 				},
+				Certificate: vault.Certificate{
+					TTL: 26 * 7 * 24 * time.Hour,
```
rossf7: This was the best I could come up with for 6 months in hours. Open to suggestions for improvement :)

installation/leaseweb.go
```diff
@@ -33,8 +33,11 @@ var Leaseweb = configuration.Installation{
 				CA: vault.CA{
 					TTL: 10 * 365 * 24 * time.Hour,
 				},
+				Certificate: vault.Certificate{
```
rossf7: I assumed we want to do this for AWS and Leaseweb. Let me know if that's incorrect.

installation/aws.go
xh3b4sd:
```
time.Hour * 24 * 30 * 6
```

rossf7: Thanks, changed :)

PR #92, opened 2017-05-31
Adds Route 53 settings to AWS provider

rossf7:
Towards giantswarm/giantswarm#1501

Adds Route53 section to the AWS provider and sets values for the hosted zones. This is because creating hosted zones should be part of installation setup.

### DIFF
--- a/installation/aws.go
+++ b/installation/aws.go
@@ -23,6 +23,8 @@ import (
 	"github.com/giantswarm/architect/configuration/provider/aws"
 	"github.com/giantswarm/architect/configuration/provider/aws/ec2"
 	"github.com/giantswarm/architect/configuration/provider/aws/ec2/instance"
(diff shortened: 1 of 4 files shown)


installation/aws.go
```diff
@@ -118,6 +120,13 @@ var AWS = configuration.Installation{
 						Default:      instance.Default,
 					},
 				},
+				Route53: route53.Route53{
+					HostedZones: hostedzones.HostedZones{
+						API:     "Z1Z5J7V0K6UO20",
+						Etcd:    "Z1Z5J7V0K6UO20",
```
rossf7: Each recordset has its own HZ ID for flexibility. We may want to make etcd a private HZ at some point.

rossf7:
Build errors fixed. I think this was VS Code getting the imports wrong and me missing it :(

PR #93, opened 2017-06-02
Add configurable intercom app id, and set it for aws and leaseweb installations

oponder:
Related to: https://github.com/giantswarm/happa/issues/158


oponder:
I'm actually not totally sure if this needs to be in architect.
It could just be set directly in Happa's kubernetes manifests.

WDYT @JosephSalisbury ?

marians:
The only thing I can review here is the Intercom ID, and that looks okay.

JosephSalisbury:
We have production and staging manifests for happa, we could set this there?

JosephSalisbury:
I don't really see this changing between installations

oponder:
Agreed, this doesn't belong in architect.

PR #94, opened 2017-06-06
Adds support for tagging and pushing latest

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1499

### DIFF
--- a/workflow/docker.go
+++ b/workflow/docker.go
@@ -12,8 +12,10 @@ var (
 	DockerRunVersionCommandName = "docker-run-version"
 	DockerRunHelpCommandName    = "docker-run-help"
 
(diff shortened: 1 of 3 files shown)


xh3b4sd:
Why do we want this? IMO we should never use `latest` due to cashing insufficiencies.

JosephSalisbury:
@rossf7 wanted it for discoverability of public images, and @calvix wanted it to ease some operational tasks

my personal opinion is that specific tags are still more reliable - latest can break api at arbitrary points, but can help with discoverability / operational ease for us

JosephSalisbury:
Basically, it's an option, people should use their best judgement :D

PR #99, opened 2017-06-16
Adds support for creating deployment events

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1253

On deployments, we create a deployment event for each environment the project is configured for. The environment name matches the environment name we configure in draughtsman.


events/events.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/google/go-github/github"
+
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+// Environment is a name of an installation.
+type Environment string
+
+var environmentProjects = map[Environment][]string{
+	"centaur": []string{
```
xh3b4sd: What is this?

JosephSalisbury: See https://gigantic.slack.com/archives/C02MS174K/p1497616415821859

xh3b4sd: Godoc would be useful. Lets describe that the environments are code names of actual G8S installations which we use to not expose real customer information.

PR #101, opened 2017-06-17
Sets required contexts empty, to allow deployment event creation in b…

JosephSalisbury:
…uild


JosephSalisbury:
ping @giantswarm/team-positive

PR #103, opened 2017-06-19
align max port range with existing Kemp port

xh3b4sd:


installation/leaseweb.go
```diff
@@ -110,7 +110,7 @@ var Leaseweb = configuration.Installation{
 				Ingress: ingress.Ingress{
 					PortRange: ingress.PortRange{
 						Min: 31000,
-						Max: 31060,
+						Max: 31021,
```
xh3b4sd: Check https://frn-lb.giantswarm.io:8443/. We only have limited IPs and defining a bigger range here would result in broken ingresses.

PR #104, opened 2017-06-20
added vikings environment

xh3b4sd:
We need this apparently for a new installation. So just getting it out here.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -35,6 +35,23 @@ var environmentProjects = map[Environment][]string{
 		"tokend",
 		"userd",
 	},
+	"vikings": []string{
+		"api",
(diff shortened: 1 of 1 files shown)


events/events.go
```diff
@@ -35,6 +35,23 @@ var environmentProjects = map[Environment][]string{
...
+		"cert-operator",
+		"cluster-service",
+		"companyd",
+		"desmotes",
+		"draughtsman",
+		"g8s-etcd-operator",
+		"g8s-prometheus",
+		"happa",
+		"kubernetesd",
+		"aws-operator",
```
xh3b4sd: It is an AWS installation, so no KVM stuff necessary here.

PR #106, opened 2017-06-23
Add NTP servers to KVM provider

rossf7:
For use in kubernetesd configmap


rossf7:
@xh3b4sd Import problem was due to VS Code and me missing it :( Going to go back to Vim until the Go extension is more stable.

rossf7:
Done :)

Issue #107, opened 2017-06-26
Add an ignore file for architect


MarcelMue:
Architect currently tries to fill all templates which it can find. This is not always welcome as templates for other applications can be inside of configmaps.

A ignore file should be added as a default in architect in order to solve this problem. This file could be named `.architect` in the `kubernetes` / `helm` folder. Architect should parse this file if it's present and ignore all files which are mentioned inside of it.

MarcelMue:
This is possibly just a helm issue - architect only handles the `kubernetes` folder, which we are getting rid of. So this would introduce a feature which is deprecated very shortly.

PR #111, opened 2017-06-28
WIP: Add ignorefile functionality for templating

MarcelMue:
This functionality is supposed to exclude files from templating in order to prevent architect from parsing go templates in `configmaps` or other yaml files which it is supposed to ignore.

- [x] Architect should accept a list of files to ignore
- [ ] Test manually if the added feature works as intended (will do so on my linux machine)
- [ ] The functionality should have at least one small test 
- [ ] Revert the `circle.yml` to the master again after reviews are accepted

towards #107


MarcelMue:
I can't easily test this right now but I'd love to have some input if this is going in the right direction. Or if I understood the workflow of architect correctly.

cmd/deploy.go
```diff
@@ -68,6 +70,8 @@ func init() {
 
 	deployCmd.Flags().StringVar(&helmDirectoryPath, "helm-directory-path", "./helm", "directory holding helm chart")
 	deployCmd.Flags().StringVar(&resourcesDirectoryPath, "resources-directory-path", "./kubernetes", "directory holding kubernetes resources")
+
+	deployCmd.Flags().StringVar(&ignorefilePath, "ignorefile-path", ".architect", "file containing a list of files to be ignored in templating")
```
kopiczko: I'd name it `.architectignore` to align with `.gitignore` or `.dockerignore`.

circle.yml
```diff
@@ -15,6 +15,6 @@ test:
 
 deployment:
     master:
-        branch: master
+        branch: add-file-ignore
```
kopiczko: Remember to revert.

template/template.go
```diff
@@ -51,7 +52,7 @@ type TemplateConfiguration struct {
 // It templates the chart's Chart.yaml and templates/deployment.yaml
 // with this information.
 // It is an error if there are multiple charts in the helm directory.
-func TemplateHelmChart(fs afero.Fs, helmPath string, buildInfo BuildInfo) error {
+func TemplateHelmChart(fs afero.Fs, helmPath string, buildInfo BuildInfo, ignorefiles []string, workingDir string) error {
```
kopiczko: Why multiple ignorefiles?

utils/ignore_files.go
```diff
@@ -0,0 +1,52 @@
...
+			ignorefiles = strings.Split(string(ignoreFile[:n]), "/n")
+		}
+		for i, ignorefile := range ignorefiles {
+			ignorefiles[i] = filepath.Join(workingDirectoryPath, ignorefile)
+		}
+	}
+	return ignorefiles, err
+}
+
+func IsInIgnorefiles(fs afero.Fs, workingDirectoryPath string, ignorefiles []string, filepathToCheck string) bool {
```
kopiczko: Did you consider https://github.com/monochromegane/go-gitignore ?

template/template.go
```diff
@@ -78,9 +78,22 @@ func TemplateHelmChart(fs afero.Fs, helmPath string, buildInfo BuildInfo, ignore
...
+		if err != nil {
+			microerror.MaskAny(err)
+		}
+
+		isIgnored := false
+		if architectignore != nil {
+			isIgnored = architectignore.Match(path, isDir)
+		}
+
+		if !isIgnored {
```
kopiczko: Returning early will allow to avoid this big indented block.

workflow/workflow.go
```diff
@@ -158,6 +161,8 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		w = append(w, dockerPushLatest)
 	}
 
+	architectIgnore, _ := gitignore.NewGitIgnore(projectInfo.ArchitectignorePath)
```
kopiczko: `func NewGitIgnore(gitignore string, base ...string) (IgnoreMatcher, error) {`

You're ignoring the returned error.

utils/ignore_files.go
```diff
@@ -0,0 +1,52 @@
...
+		}
+		n := bytes.IndexByte(ignoreFile, 0)
+		if n > 0 {
+			ignorefiles = strings.Split(string(ignoreFile[:n]), "/n")
+		}
+		for i, ignorefile := range ignorefiles {
+			ignorefiles[i] = filepath.Join(workingDirectoryPath, ignorefile)
+		}
+	}
+	return ignorefiles, err
```
kopiczko: I don't know if architect uses masked errors. I see it doesn't in `workflow.go`.

JosephSalisbury: We should be

circle.yml
JosephSalisbury: Why do we have this?

utils/ignore_files.go
kopiczko: It isn't in the scope of this change, though. I just highlighted it.

circle.yml
MarcelMue: This is so that a build is not marked as the latest master build but the latest build of the branch.

The latest master branch would be used in builds afaik

JosephSalisbury: https://circleci.com/gh/giantswarm/architect/333

If you look at the build, we've been using this branch's version as our architect version for all builds. I would revert this now.

MarcelMue: You are correct, the build is actually tagged as latest as well. Not sure what the branch even does then.

MarcelMue:
Had a chat with @JosephSalisbury :
This PR is not the real issue, as it would only fix the templating done by architect in the `kubernetes` folder. Templating is of course also done by helm later on, which isn't changed by this PR.

Since we are getting rid of `kubernetes` folders all together, this PR would introduce a feature that is no longer needed.

Issue #73, opened 2017-05-24
Integrate architect templating into workflow


JosephSalisbury:
The templating happens outside of the workflow, which is a bit magic.

If we refactor commands into an interface, we should be able to run arbitrary templating commands as part of the workflow, which should clean things up a lot.

JosephSalisbury:
There's probably more things than just templating that we can refactor behind a new command interface, have a look for that too.

JosephSalisbury:
Only really makes sense after https://github.com/giantswarm/giantswarm/issues/1474

JosephSalisbury:
Picking this up as a getting-back-to-work starter task, and to make https://github.com/giantswarm/giantswarm/issues/1606 easier

PR #113, opened 2017-07-10
Refactors commands structs into tasks interface

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/73

- Removed `commands` package
- Adds `tasks` package, which introduces a `Task` interface
- Adds `ExecTask` implementation - essentially `commands.Command`, but implementing `tasks.Task`
` Refactors current `commands.Command` to use the `Task` interface

Next PR I am going to refactor templating to use `tasks.Task`


tasks/spec.go
```diff
@@ -0,0 +1,27 @@
+package tasks
+
+import (
+	"log"
+
+	microerror "github.com/giantswarm/microkit/error"
+)
+
+// Task represents a piece of work to perform.
+type Task interface {
```
JosephSalisbury: This is the magic. Instead of everything having to boil down to a `command.Command` struct, we implement `Task` instead. This means we can also move templating and event creation into the workflow, which will really clean stuff up.

workflow/helm.go
```diff
@@ -53,29 +54,29 @@ func NewHelmLoginCommand(fs afero.Fs, projectInfo ProjectInfo) (commands.Command
...
-		return commands.Command{}, err
+		return nil, err
 	}
 	if !helmDirExists {
-		return commands.Command{}, noHelmDirectoryError
+		return nil, noHelmDirectoryError
 	}
 
-	cndDir, err := cnrDirectory()
+	cnrDir, err := cnrDirectory()
```
JosephSalisbury: Ninja fix.

tasks/spec.go
```diff
@@ -0,0 +1,27 @@
...
+	Name() string
+	Run() error
+	String() string
```
kopiczko: If this is to enforce implementing https://golang.org/pkg/fmt/#Stringer I'd embed `Stringer`.

tasks/spec.go
kopiczko: Is it naming issue? Would you achieve the same by converting current `Command` struct to interface like this?

JosephSalisbury: I kind of hated having `Command`, as it means we had packages named `cmd` (for cobra), and `command`, and it consistently confused me. Hence the new name.

JosephSalisbury: So, to avoid personal confusion, mainly :D

tasks/spec.go
JosephSalisbury: Yeah, I like that

JosephSalisbury: Changed

PR #116, opened 2017-07-11
Refactors kubernetes resource templating to task interface

JosephSalisbury:
Towards https://github.com/giantswarm/architect/issues/73

### DIFF
--- a/template/error.go
+++ b/template/error.go
@@ -45,3 +45,10 @@ var incorrectTemplatingError = errgo.New("incorrect templating")
 func IsIncorrectTemplating(err error) bool {
 	return errgo.Cause(err) == incorrectTemplatingError
 }
+
+var emptyKubernetesResourcesDirectoryPath = errgo.New("empty kubernetes resources directory path")
+
+// IsEmptyKubernetesResourcesDirectoryPath asserts emptyKubernetesResourcesDirectoryPath
+func IsEmptyKubernetesResourcesDirectoryPath(err error) bool {
+	return errgo.Cause(err) == emptyKubernetesResourcesDirectoryPath
+}
--- a/template/kubernetes.go
+++ b/template/kubernetes.go
@@ -1,0 +1,136 @@
+package template
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/spf13/afero"
+
+	microerror "github.com/giantswarm/microkit/error"
+
+	"github.com/giantswarm/architect/configuration"
+	"github.com/giantswarm/architect/utils"
+)
+
+const (
+	TemplateKubernetesResourcesTaskName = "template-kubernetes-resources"
+
+	// TemplateKubernetesResourcesTaskString is the format for printing the
+	// kubernetes resources templating task.
+	// Name of the task, templated resources directory path, sha.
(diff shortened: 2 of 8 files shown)


template/error.go
```diff
@@ -45,3 +45,10 @@ var incorrectTemplatingError = errgo.New("incorrect templating")
 func IsIncorrectTemplating(err error) bool {
 	return errgo.Cause(err) == incorrectTemplatingError
 }
+
+var emptyKubernetesResourcesDirectoryPath = errgo.New("empty kubernetes resources directory path")
```
xh3b4sd: `emptyDirectoryPathError` would work for me. You could then mask the error with additional information ala `microerror.MaskAnyf(emptyDirectoryError, "kubernetes resource '%s'", thePath)`.

template/kubernetes.go
```diff
@@ -0,0 +1,136 @@
...
+func (t TemplateKubernetesResourcesTask) Run() error {
+	if t.kubernetesResourcesDirectoryPath == "" {
+		return microerror.MaskAny(emptyKubernetesResourcesDirectoryPath)
+	}
+
+	if err := utils.CopyDir(
+		t.fs,
+		t.kubernetesResourcesDirectoryPath,
+		t.templatedResourcesDirectoryPath,
+	); err != nil {
```
xh3b4sd: Every time function calls and conditionals are mixed a pomsky puppy dies. Please don't kill the puppies.

template/kubernetes.go
```diff
@@ -0,0 +1,136 @@
...
+// templateKubernetesResources takes a filesystem,
+// a path to a directory holding kubernetes resources,
+// and an installation configuration.
+// It templates the given resources, with data from the configuration,
+// writing changes to the files.
+func templateKubernetesResources(fs afero.Fs, resourcesPath string, config TemplateConfiguration) error {
+	walkFunc := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			microerror.MaskAny(err)
+		}
```
xh3b4sd: I did this myself recently and wondered what the use case for this propagated error is. This feels super weird.

template/kubernetes.go
```diff
@@ -0,0 +1,136 @@
...
+
+		var buf bytes.Buffer
+		if err := t.Execute(&buf, config); err != nil {
+			microerror.MaskAny(err)
+		}
+
+		templatedContents := buf.String()
+
+		// This add backwards compatability for `%%DOCKER_TAG%%`. Deprecated.
```
xh3b4sd: He, she it, das "S" muss mit. :D

template/kubernetes.go
JosephSalisbury: I know we disagree with
```
if err := func(); err != nil {
```

gonna go with this :D

template/kubernetes.go
JosephSalisbury: I've never even seen this error actually be anything except for `nil`. Feels wrong not to handle it, but I don't think it's a big deal.

template/kubernetes.go
JosephSalisbury: ja ja genau

JosephSalisbury: fixeden

template/error.go
JosephSalisbury: no strong feelings. this error will go pretty soon when we remove the push deployments support

PR #118, opened 2017-07-11
Changes default registry to quay

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1458

All of the boyscouting! :D


cmd/deploy.go
```diff
@@ -49,15 +49,9 @@ func init() {
 	var defaultDockerPassword string
 
 	if os.Getenv("CIRCLECI") == "true" {
```
xh3b4sd: Unrelated: I saw Timo using `architect` locally to build projects. Why should we only support these env vars on Circle? We can also use them locally.

JosephSalisbury: You can use `architect build` locally, it works pretty nice. `architect deploy` is a bit funny. We can sort this out as we go.

PR #127, opened 2017-07-17
Adds support for Helm templating in deployment.yaml and Chart.yaml

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1655

We require support for templating deployment.yaml files with the
latest image tag, as well as having Helm templating inside these
files - see https://github.com/giantswarm/api/pull/420

This introduces a method that the architect templating uses, which
allows for templating the BuildInfo struct, without clobbering the
Helm templating, as well as being open to new fields in the future.

### DIFF
--- a/template/helm.go
+++ b/template/helm.go
@@ -59,20 +59,23 @@ func (t TemplateHelmChartTask) Run() error {
 	for _, path := range paths {
 		exists, err := afero.Exists(t.fs, path)
 		if err != nil {
-			microerror.MaskAny(err)
+			return microerror.MaskAny(err)
 		}
 
 		if exists {
 			contents, err := afero.ReadFile(t.fs, path)
 			if err != nil {
-				microerror.MaskAny(err)
+				return microerror.MaskAny(err)
 			}
 
 			buildInfo := BuildInfo{SHA: t.sha}
 			templatedContents, err := SafeTemplate(buildInfo, contents)
+			if err != nil {
+				return microerror.MaskAny(err)
+			}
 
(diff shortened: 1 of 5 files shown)


kopiczko:
FYI: reviewing

template/helm.go
```diff
@@ -71,18 +69,9 @@ func (t TemplateHelmChartTask) Run() error {
 			}
 
 			buildInfo := BuildInfo{SHA: t.sha}
+			templatedContents, err := SafeTemplate(buildInfo, contents)
```
kopiczko: error not handled

xh3b4sd: This error is not handled.

template/safe.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	// TemplateFieldFormat is the format for fields that SafeTemplate will template.
+	TemplateFieldFormat = "{{ .%s }}"
+
+	// StringTypeName is the name of the string type.
+	StringTypeName = "string"
+)
+
+// safeTemplate takes a struct and a template byte array, and attempts to
```
xh3b4sd: The method is public with an upper S.

JosephSalisbury:
Not certain if using `reflect` was a good idea, but I didn't want to have two places where we need to define `BuildInfo` fields.

JosephSalisbury:
We need to support helm templating in `deployment.yaml`. The current architect templating completely clobbers it, so we need to handle it a bit 'smarter', for various definitions of smarter.

JosephSalisbury:
Using different delims is a better idea. Going with that for now.

PR #131, opened 2017-07-31
Add ingress controller version to the guest provider

rossf7:
Towards giantswarm/giantswarm#1695

This PR adds the version of the Ingress Controller docker image to the guest provider. This is so we use a retagged image and can replace the hardcoded image in `k8scloudconfig`.


rossf7:
@xh3b4sd Ack. I've added `installations` to my list.

PR #132, opened 2017-07-31
update-kubectl-version

calvix:
safe me ass


calvix:
so much pain because of tests :)

PR #142, opened 2017-08-11
added retry task

xh3b4sd:


tasks/retry.go
```diff
@@ -0,0 +1,52 @@
...
+	err := backoff.RetryNotify(o, t.BackOff, n)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (t RetryTask) Name() string {
+	return t.Task.Name()
```
JosephSalisbury: It makes sense to me for this to be something like:

```
return fmt.Sprintf("retry task: %v", t.Task.Name())
```

so it's clear that it's a retry task, but also the underlying task is clear.

tasks/retry.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (t RetryTask) String() string {
+	return t.Task.String()
```
JosephSalisbury: Similar to here

PR #146, opened 2017-08-16
Adds retry pull tasks for golang and helm

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1785

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -73,6 +73,16 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		}
 	}
 	if goLangFilesExist {
+		{
+			golangPull, err := NewGoPullTask(fs, projectInfo)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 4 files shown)


workflow/workflow.go
```diff
@@ -141,9 +151,6 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 				return nil, err
 			}
 			wrappedDockerLogin := tasks.NewRetryTask(backoff.NewExponentialBackOff(), dockerLogin)
-			if err != nil {
```
JosephSalisbury: Ninja, these were just checking the previous error again.

workflow/workflow.go
```diff
@@ -73,6 +73,16 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		}
 	}
 	if goLangFilesExist {
+		{
+			golangPull, err := NewGoPullTask(fs, projectInfo)
+			if err != nil {
+				return nil, err
```
xh3b4sd: The error masking is still missing in the whole project.

xh3b4sd:
Note that CI is breaking.

PR #148, opened 2017-08-17
Add archon

r7vme:


events/events.go
```diff
@@ -203,6 +203,25 @@ var environmentProjects = map[Environment][]string{
...
+		"g8s-grafana",
+		"g8s-prometheus",
+		"happa",
+		"kubernetesd",
+		"aws-operator",
+		"passage",
+		"testbot",
+		"tokend",
+		"userd",
+	},
```
kopiczko: does it make sense to extract to `var`s here? one with `aws` set and another one with `kvm` set?

ping @JosephSalisbury

xh3b4sd: We are going to remove this anyway soon. So any refactoring love is nuts IMO. This was always only mend to be temporary.

PR #149, opened 2017-08-21
Branch name

xh3b4sd:
Fixes https://github.com/giantswarm/giantswarm/issues/1605.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -81,7 +81,9 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		WorkingDirectory: workingDirectory,
 		Organisation:     organisation,
 		Project:          project,
-		Sha:              sha,
+
+		Branch: branch,
+		Sha:    sha,
 
 		Registry:       registry,
 		DockerEmail:    dockerEmail,
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,7 +20,9 @@ var (
 	registry     string
 	organisation string
 	project      string
-	sha          string
+
+	branch string
+	sha    string
 
 	dryRun bool
 )
@@ -46,17 +48,32 @@ func init() {
 	}
 
 	// We also use the git HEAD commit sha as well.
-	out, err := exec.Command("git", "rev-parse", "HEAD").Output()
-	if err != nil {
-		log.Fatalf("could not get git sha: %v\n", err)
+	var defaultSha string
+	{
+		out, err := exec.Command("git", "rev-parse", "HEAD").Output()
+		if err != nil {
+			log.Fatalf("could not get git sha: %#v\n", err)
+		}
+		defaultSha = strings.TrimSpace(string(out))
 	}
-	defaultSha := strings.TrimSpace(string(out))
+
+	// We also use the git HEAD branch as well.
+	var defaultBranch string
+	{
+		out, err := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD").Output()
+		if err != nil {
+			log.Fatalf("could not get git branch: %#v\n", err)
+		}
+		defaultBranch = strings.TrimSpace(string(out))
+	}
 
 	RootCmd.PersistentFlags().StringVar(&workingDirectory, "working-directory", defaultWorkingDirectory, "working directory for architect")
 
 	RootCmd.PersistentFlags().StringVar(&registry, "registry", "quay.io", "docker registry")
 	RootCmd.PersistentFlags().StringVar(&organisation, "organisation", defaultOrganisation, "organisation who owns the project")
 	RootCmd.PersistentFlags().StringVar(&project, "project", defaultProject, "name of the project")
+
+	RootCmd.PersistentFlags().StringVar(&branch, "branch", defaultBranch, "git branch to build")
 	RootCmd.PersistentFlags().StringVar(&sha, "sha", defaultSha, "git SHA1 to build")
 
 	RootCmd.PersistentFlags().BoolVar(&dryRun, "dry-run", dryRun, "show what would be executed, but take no action")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,7 +12,7 @@ import (
 var (
 	RootCmd = &cobra.Command{
 		Use:   "architect",
-		Short: "Architect is a tool for managing builds within the Giant Swarm releases engineering.",
+		Short: "Architect is a tool for managing builds within Giant Swarm release engineering.",
 	}
 
 	workingDirectory string
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,7 +12,7 @@ import (
 var (
 	RootCmd = &cobra.Command{
 		Use:   "architect",
-		Short: "architect is a tool for managing Giant Swarm release engineering",
+		Short: "Architect is a tool for managing builds within the Giant Swarm releases engineering.",
 	}
 
 	workingDirectory string
--- a/workflow/docker.go
+++ b/workflow/docker.go
@@ -16,6 +16,14 @@ const (
 	DockerTagLatestTaskName  = "docker-tag-latest"
 	DockerPushShaTaskName    = "docker-push-sha"
 	DockerPushLatestTaskName = "docker-push-latest"
+
+	// DockerImageRefFmt is the format string used to compute the reference of the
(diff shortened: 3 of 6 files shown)


cmd/deploy.go
```diff
@@ -81,7 +81,9 @@ func runDeploy(cmd *cobra.Command, args []string) {
 		WorkingDirectory: workingDirectory,
 		Organisation:     organisation,
 		Project:          project,
-		Sha:              sha,
+
+		Branch: branch,
+		Sha:    sha,
```
xh3b4sd: Here the branch is additionally set in the project info which the sub commands receive.

cmd/root.go
```diff
@@ -12,15 +12,17 @@ import (
 var (
 	RootCmd = &cobra.Command{
 		Use:   "architect",
-		Short: "architect is a tool for managing Giant Swarm release engineering",
+		Short: "Architect is a tool for managing builds within the Giant Swarm releases engineering.",
```
xh3b4sd: Boyscouting here. I think this is more right.

cmd/root.go
```diff
@@ -46,17 +48,32 @@ func init() {
...
 	}
-	defaultSha := strings.TrimSpace(string(out))
 
 	RootCmd.PersistentFlags().StringVar(&workingDirectory, "working-directory", defaultWorkingDirectory, "working directory for architect")
 
 	RootCmd.PersistentFlags().StringVar(&registry, "registry", "quay.io", "docker registry")
 	RootCmd.PersistentFlags().StringVar(&organisation, "organisation", defaultOrganisation, "organisation who owns the project")
 	RootCmd.PersistentFlags().StringVar(&project, "project", defaultProject, "name of the project")
+
+	RootCmd.PersistentFlags().StringVar(&branch, "branch", defaultBranch, "git branch to build")
```
xh3b4sd: The branch is configurable via flag just like the other stuff.

workflow/docker.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-var (
+const (
```
xh3b4sd: Boyscouting I think having these immutable as constants is more right.

workflow/docker.go
```diff
@@ -16,6 +16,14 @@ var (
...
 	DockerPushShaTaskName    = "docker-push-sha"
 	DockerPushLatestTaskName = "docker-push-latest"
+
+	// DockerImageRefFmt is the format string used to compute the reference of the
+	// Docker image used to build and push. It may look something like this.
+	//
+	//     quay.io/giantswarm/architect:master-e8363ac222255e991c126abe6673cd0f33934ac8
+	//
+	DockerImageRefFmt    = "%s/%s/%s:%s-%s"
+	LatestDockerImageTag = "latest"
```
xh3b4sd: We use now constants and the format is documented and uses proper format placeholders. Please stop using `%v` for everything.

workflow/docker.go
```diff
@@ -50,13 +58,7 @@ func NewDockerBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error
...
 			"build",
 			"-t",
-			fmt.Sprintf(
-				"%v/%v/%v:%v",
-				projectInfo.Registry,
-				projectInfo.Organisation,
-				projectInfo.Project,
-				projectInfo.Sha,
-			),
+			newDockerImageRef(projectInfo, projectInfo.Sha),
```
xh3b4sd: A helper function makes this code a bit more DRY and clean.

workflow/docker.go
```diff
@@ -72,14 +74,9 @@ func NewDockerRunVersionTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task,
...
-				"%v/%v/%v:%v",
-				projectInfo.Registry,
-				projectInfo.Organisation,
-				projectInfo.Project,
-				projectInfo.Sha,
-			),
-			Args: []string{"version"},
+			Args:             []string{"version"},
+			Image:            newDockerImageRef(projectInfo, projectInfo.Sha),
+			WorkingDirectory: projectInfo.WorkingDirectory,
```
xh3b4sd: I am not sure if I am breaking something now by fixing it. The working directory was empty before. The tests I wrote showed that. Should we be worried now?

workflow/docker_test.go
```diff
@@ -0,0 +1,149 @@
+package workflow
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/giantswarm/architect/tasks"
+)
+
+func Test_Workflow_Docker_NewDockerBuildTask(t *testing.T) {
```
xh3b4sd: Here are actually some tests to make sure the args are right. LGTM.

workflow/docker_test.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func testNewProjectInfo() ProjectInfo {
+	return ProjectInfo{
+		WorkingDirectory: "/usr/code/",
+		Organisation:     "giantswarm",
+		Project:          "architect",
+
+		Branch: "master",
+		Sha:    "e8363ac222255e991c126abe6673cd0f33934ac8",
```
xh3b4sd: I chose more real settings here to make the result more readable. It was super hard to parse for me using strings like `test-sha`. This data is more aligned with the real world.

workflow/workflow.go
```diff
@@ -30,7 +30,9 @@ type ProjectInfo struct {
 	WorkingDirectory string
 	Organisation     string
 	Project          string
-	Sha              string
+
+	Branch string
+	Sha    string
```
xh3b4sd: Here is the branch actually added to the `ProjectInfo` type.

cmd/root.go
rossf7: "within Giant Swarm release engineering" reads more easily for me.

workflow/docker.go
rossf7: <3

workflow/docker.go
rossf7: I don't know architect well enough to comment. @JosephSalisbury WDYT?

cmd/root.go
xh3b4sd: Sweet. Done.

workflow/docker.go
```diff
@@ -204,14 +179,20 @@ func NewDockerPushLatestTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task,
...
-				projectInfo.Project,
-			),
+			newDockerImageRef(projectInfo, LatestDockerImageTag),
 		},
 	)
 
 	return dockerPush, nil
 }
+
+func newDockerImageRef(projectInfo ProjectInfo, version string) string {
```
JosephSalisbury: I'd like to see a test over this.

workflow/docker.go
JosephSalisbury: I'm actually not certain here. If architect builds fine locally, then I don't see what else it could break.

JosephSalisbury:
I would recommend trying to deploy, end to end, with the new tag scheme. "It should work" probably isn't good enough there, and we should make sure before merging this.

workflow/docker.go
xh3b4sd: I think this is already tested multiple times. Anyway added an explicit test.

workflow/docker.go
xh3b4sd: I could build architect locally with architect. Looks like this is fine.

xh3b4sd:
> I would recommend trying to deploy, end to end, with the new tag scheme.

How to do this without merging it? I miss something here.

JosephSalisbury:
Grab a helm chart for some service from quay, modify the version, push it, create a deployment event for that version.

xh3b4sd:
I cannot test this locally. I cannot install helm and the registry plugin. The latest version is broken since a week. Don't know how to install another version. Need a pair on this.

xh3b4sd:
For the record. I am removing the branch prefixing and keep the tests and stuff. See https://github.com/giantswarm/giantswarm/issues/1605#issuecomment-324326336.

xh3b4sd:
Please have another look guys. I removed the branch prefix in the tags. I would like to keep the branch information within architect though. I do not want to waste time on cleaning this up and I think this information can be useful for logging, debugging, meta data at some point.

PR #152, opened 2017-08-24
fix build command by setting docker credentials

xh3b4sd:


cmd/deploy.go
```diff
@@ -23,10 +23,6 @@ var (
 		Run:   runDeploy,
 	}
 
-	dockerEmail    string
```
JosephSalisbury: We still need these for the docker tag latest, no?

xh3b4sd:
@oponder show me how to write such a test and I will do it.

cmd/deploy.go
xh3b4sd: This is because of the shitty design of the command line tool itself. Everything is in the same package and everything uses the same vars. These vars are used in both build and deploy commands. I just moved the vars to the file where the command is defined which uses the vars first.

xh3b4sd: I would like to fix this by refactoring architect to use a project and command package structure like in opsctl. But I do not want to do this now, because there is no real benefit right now. So some boyscouting left to do which I am not fuzzed about now.

oponder:
@xh3b4sd I have no clue :P

xh3b4sd:
Same here. Then lets go.

PR #174, opened 2017-11-16
promote chart to stable channel on deploy

fgimenez:
These changes add the just built chart to the `stable` channel on deploy, this will allow us to reference and use the charts currently deployed from CI setups.

In order to make unit tests happy the promotion to stable is only done when a dockerfile is in place. Also the promotion task is bound to the stable channel, imo we don't need to parameterize the channel, let me know wdyt.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -247,11 +247,14 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 			}
 			w = append(w, helmLogin)
 
-			helmPushToChannel, err := NewHelmPromoteToStableChannelTask(fs, projectInfo)
+			helmPromoteToChannel, err := NewHelmPromoteToStableChannelTask(fs, projectInfo)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


workflow/workflow.go
```diff
@@ -219,6 +219,41 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
...
+			}
+			w = append(w, helmChartTemplate)
+
+			helmLogin, err := NewHelmLoginTask(fs, projectInfo)
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+			w = append(w, helmLogin)
+
+			helmPushToChannel, err := NewHelmPromoteToStableChannelTask(fs, projectInfo)
```
xh3b4sd: This task should be wrapped by a retry task as well. In case there is a network problem on Circle, which we see quite often, the task is retried and the build succeeds, even when taking a bit longer. At least no human action is required then to get the build done

fgimenez: Nice! thx, done

PR #178, opened 2017-11-29
Walk to render helm charts

MarcelMue:
I am unsure if this will cause any problems, the idea here is to walk the `*-chart` folder so we don't have specific files which are rendered anymore.

I used `afero.Walk` here but am a bit unsure if this is the right way to go.


MarcelMue:
I added a file to the existing test that the walk through different filenames works now. 
The change simply causes the existing render logic to not only do two static files but walk the entire folder. Are there any other sensible tests for this?

MarcelMue:
Can you and especially @JosephSalisbury  take another look ?
I am blocked by this.

xh3b4sd:
I am still wondering why you changed an existing test instead of adding another one. When testing such algorithms it is always good to have tests to verify work. Now you changed the code and changed the tests. How do we know the old tests still work?

MarcelMue:
Okay, I get your issue now. I have removed changes that made it seem like existing tests did not run. The diff should now show that I added some new cases while the old ones are still good.

PR #184, opened 2018-01-19
Add pv-cleaner-operator for centaur

corest:


xh3b4sd:
What is the state of this? Can we merge? Is this still relevant? What is holding us back?

PR #188, opened 2018-02-22
Add g8s-efk to all installations

marians:
This would allow deployment events for g8s-efk for all installations.

**This should not be merged before we manually deployed everywhere.**


marians:
For https://github.com/giantswarm/giantswarm/issues/1877

marians:
By now we have manually deployed to all installations, without causing problems it seems. Merging.

PR #189, opened 2018-02-22
add multiple chart support

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2520


template/helm.go
```diff
@@ -32,29 +31,14 @@ const (
 
 type TemplateHelmChartTask struct {
 	fs       afero.Fs
-	helmPath string
+	chartDir string
 	sha      string
 }
 
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 // It is an error if there are multiple charts in the helm directory.
```
fgimenez: This comment should be removed.

template/helm_test.go
```diff
@@ -13,117 +13,87 @@ import (
...
-				if err != nil {
-					return microerror.Mask(err)
-				}
-				if len(fileInfos) != 0 {
-					return microerror.Mask(multipleHelmChartsError)
-				}
-
-				return nil
-			},
-		},
```
kopiczko: Correct me if I'm wrong but this test tests afero implementation instead of NewTemplateHelmChartTask

workflow/workflow.go
```diff
@@ -204,11 +229,13 @@ func NewDeploy(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
...
-		dockerTagLatest, err := NewDockerTagLatestTask(fs, projectInfo)
-		if err != nil {
-			return nil, microerror.Mask(err)
+		{
+			dockerTagLatest, err := NewDockerTagLatestTask(fs, projectInfo)
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+			w = append(w, dockerTagLatest)
 		}
```
kopiczko: Just moved to a scope like everything else here.

workflow/workflow.go
```diff
@@ -165,23 +166,46 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
...
+		fileInfos, err := afero.ReadDir(fs, helmDirectory)
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
-		w = append(w, helmChartTemplate)
 
-		helmLogin, err := NewHelmLoginTask(fs, projectInfo)
-		if err != nil {
-			return nil, microerror.Mask(err)
+		if len(fileInfos) > 0 {
```
fgimenez: Why not returning early if len(fileInfos) == 0?

kopiczko: This is how this function is designed. There is a possibility some add new `if xxxExists {` and that would be like a trap. Out of the sudden architect stops execution when helm directory is empty. It would be a bug very difficult to find.

workflow/workflow.go
```diff
@@ -165,23 +166,46 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
...
+			if !strings.HasPrefix(fileInfo.Name(), prefix) {
+				return nil, microerror.Maskf(invalidHelmDirectoryError, "%q must start with %q", fileInfo.Name(), prefix)
+			}
+			if !strings.HasSuffix(fileInfo.Name(), suffix) {
+				return nil, microerror.Maskf(invalidHelmDirectoryError, "%q must end with %q", fileInfo.Name(), suffix)
+			}
+
+			chartDir := filepath.Join(helmDirectory, fileInfo.Name())
+
+			helmChartTemplate, err := NewTemplateHelmChartTask(fs, chartDir, projectInfo)
```
MarcelMue: How about adding a test for this new logic in `workflow_test.go`. There are some test cases that test deploy.

kopiczko: I added a lot of tests.

PR #190, opened 2018-02-26
add publish command

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

Adds a `publish` command to push the project chart to the specified channels (beta and testing by default). The push task is wrapped in a retry.


rossf7:
That's great that new channels can be created via the flag.

PR #191, opened 2018-02-27
do not push empty channels

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/1902

Prevents trying to push an empty string as channel name.

### DIFF
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -300,7 +300,7 @@ func NewPublish(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 	if chartDirectoryExists {
 		for _, c := range projectInfo.Channels {
 			if c == "" {
-				continue
+				return nil, microerror.Mask(emptyChannelError)
 			}
 			helmPromoteToChannel, err := NewHelmPromoteToChannelTask(fs, chartDirectory, projectInfo, c)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


workflow/workflow.go
```diff
@@ -299,6 +299,9 @@ func NewPublish(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 
 	if chartDirectoryExists {
 		for _, c := range projectInfo.Channels {
+			if c == "" {
+				continue
+			}
```
kopiczko: Why would we have empty channel names? Isn't that input parameters validation issue?

fgimenez: > Why would we have empty channel names?

If we pass something like `--channels beta,,stable`

> Isn't that input parameters validation issue?

It can be seen like that, currently we don't have actual input parameters validation logic. Do you think we should add it and fail if we pass empty channels?

xh3b4sd: I think we should just validate the input and fail in case we get something like `--channels beta,,stable`. Input like this indicates for me that something is wrong and we should be aware of it.

kopiczko: I feel the same. I'm just afraid something will fail silently in the future.

fgimenez: Makes sense, I've updated it to fail on empty channels received, PTAL.

PR #203, opened 2018-03-13
uses caching for golang artifacts

xh3b4sd:
Towards https://github.com/giantswarm/aws-operator/issues/837.

### DIFF
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -152,9 +152,11 @@ func NewGoVetTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 					projectInfo.Organisation,
 					projectInfo.Project,
 				),
+				"/tmp/go/cache:/go/cache",
 			},
 			Env: []string{
 				"GOPATH=/go",
+				"GOCACHE=/go/cache",
 			},
 			WorkingDirectory: fmt.Sprintf(
 				"/go/src/github.com/%v/%v",
@@ -184,10 +186,12 @@ func NewGoTestTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 					projectInfo.Organisation,
 					projectInfo.Project,
 				),
+				"/tmp/go/cache:/go/cache",
 			},
 			Env: []string{
 				fmt.Sprintf("GOOS=%v", projectInfo.Goos),
 				fmt.Sprintf("GOARCH=%v", projectInfo.Goarch),
+				"GOCACHE=/go/cache",
 				"GOPATH=/go",
 				"CGOENABLED=0",
 			},
@@ -219,10 +223,12 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 					projectInfo.Organisation,
 					projectInfo.Project,
 				),
+				"/tmp/go/cache:/go/cache",
 			},
 			Env: []string{
 				fmt.Sprintf("GOOS=%v", projectInfo.Goos),
 				fmt.Sprintf("GOARCH=%v", projectInfo.Goarch),
+				"GOCACHE=/go/cache",
 				"GOPATH=/go",
 				"CGOENABLED=0",
 			},
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -81,18 +81,6 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		}
 		w = append(w, goFmt)
 
-		goVet, err := NewGoVetTask(fs, projectInfo)
-		if err != nil {
-			return nil, microerror.Mask(err)
-		}
-		w = append(w, goVet)
-
-		goTest, err := NewGoTestTask(fs, projectInfo)
-		if err != nil {
-			return nil, microerror.Mask(err)
-		}
-		w = append(w, goTest)
-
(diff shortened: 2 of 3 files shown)


workflow/golang.go
```diff
@@ -152,9 +152,11 @@ func NewGoVetTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 					projectInfo.Organisation,
 					projectInfo.Project,
 				),
+				"/tmp/go/cache:/go/cache",
```
xh3b4sd: We now mount an additional volume for the configured cache dir. That way we reuse the cache across container runs during the same build. Since we build every build from scratch there shouldn't be issues with reusing old or wrong cache artefacts.

workflow/golang.go
```diff
@@ -152,9 +152,11 @@ func NewGoVetTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
...
 			},
 			Env: []string{
 				"GOPATH=/go",
+				"GOCACHE=/go/cache",
```
xh3b4sd: This defines the cache dir.

workflow/workflow.go
```diff
@@ -81,18 +81,6 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		}
 		w = append(w, goFmt)
 
-		goVet, err := NewGoVetTask(fs, projectInfo)
-		if err != nil {
-			return nil, microerror.Mask(err)
-		}
-		w = append(w, goVet)
```
xh3b4sd: `go vet` is removed since now `go test` runs `go vet` to a certain extend. We have to decide if this is good enough though. I am fine with this. The time advantage is way bigger than certain checks we might not run during the testing phase. 

> As part of building a test binary, go test runs go vet on the package and its test source files to identify significant problems. If go vet finds any problems, go test reports those and does not run the test binary. Only a high-confidence subset of the default go vet checks are used.

See https://golang.org/cmd/go/#hdr-Test_packages.

workflow/workflow.go
```diff
@@ -104,6 +92,12 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 			}
 			w = append(w, goBuild)
 		}
+
+		goTest, err := NewGoTestTask(fs, projectInfo)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		w = append(w, goTest)
```
xh3b4sd: `go build` is now run before the tests in order to speed up the build even more. My testing showed that there is quite an advantage in terms of caching and thus execution time. I cannot really explain why this is so but my results where obvious.

workflow/workflow_test.go
```diff
@@ -106,9 +106,8 @@ func TestGetBuildWorkflow(t *testing.T) {
 			expectedTaskNames: []string{
 				GoPullTaskName,
 				GoFmtTaskName,
-				GoVetTaskName,
-				GoTestTaskName,
 				GoBuildTaskName,
+				GoTestTaskName,
```
xh3b4sd: Just aligning the tests with my changes here.

workflow/workflow.go
kopiczko: While I 100% like using caching, this is controversial (and out of scope of this PR title).

Will it cover format string parameters mismatch like `fmt.Printf("%s")`?

xh3b4sd: So the caching used here is nothing else than on your local machine. IMO we would have had problems already when there were any. And yes, `go test` catches wrong format string parameters. I already had this case.

kopiczko: I still have mixed feelings. Was `go vet` slow?

fgimenez: I agree this should be proposed in a separate PR, it's not related to the rest of the changes.

workflow/golang.go
fgimenez: Could you please explain how this env var is used? Does `go {build,test}` take it into account?

workflow/workflow.go
xh3b4sd: `go vet` is super slow. Getting rid of it safes like 60% of the time savings we gain in this PR.

xh3b4sd: I feel like the proposed changes want to do one thing. Improve build times. There changed 20 lines. Why should there be different PRs when this is about one and the same thing? It is not like I need to draw a big picture and prepare a story and migration path. This change here is how it should be based on my current point of view. I miss some constructive feedback. So what can we do to move forward? Should I change something? Can we go with this? Do we want to drop this idea?

xh3b4sd:
I want to have two more approvals here.

workflow/workflow.go
rossf7: It's a shame we need to remove go vet but at the moment I think the priority should be getting build times down.

I'm fine with the change being in this PR but I think the title should be changed for the history.  Removing `go vet` is a significant change.

Do we also want an issue to re-enable once we've deleted version sets in `aws-operator`?

xh3b4sd: It is not that `go vet` is not run at all. `go test` runs certain parts of it and this support will be extended in the future where appropriate. So I am not fuzzed about missing some weaker parts of `go vet` at all. I do not go back the older state when this here gets through. I am fine with the `go test` support personally.

xh3b4sd:
One more approve please.

xh3b4sd:
Sweet. Options everywhere. Nice.

PR #204, opened 2018-03-14
Add concurrent tasks and run go tasks concurrently

fgimenez:
Towards https://github.com/giantswarm/giantswarm/issues/2711 and https://github.com/giantswarm/giantswarm/issues/2678

Adds a `ConcurrentTask` type that runs a set of tasks concurrently using `errgroup.Group`. Also introduces changes to run the Go tasks concurrently, this is compatible with the changes introduced in #203, if we want to reuse the cache from the build step we just need to remove the build task from the concurrent set and execute it in the first place. We can measure in which situation we can have more gains (run build and test sequentially and take advantage of build cache in test or run build and test concurrently without cache).

This concurrent task can be used in more workflows, if we agree on the approach I'll propose more changes next.

### DIFF
--- a/tasks/concurrent.go
+++ b/tasks/concurrent.go
@@ -1,0 +1,51 @@
+package tasks
+
+import (
+	"bytes"
+
+	"github.com/giantswarm/microerror"
+	"golang.org/x/sync/errgroup"
+)
+
+type ConcurrentTask struct {
+	name  string
+	tasks []Task
+}
+
+func (c *ConcurrentTask) Run() error {
+	var g errgroup.Group
+
+	for _, t := range c.tasks {
+		g.Go(t.Run)
+	}
+
+	err := g.Wait()
+
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (c *ConcurrentTask) Name() string {
+	return c.String()
+}
+
+func (c *ConcurrentTask) String() string {
+	var buffer bytes.Buffer
+
+	for _, t := range c.tasks {
+		buffer.WriteString(t.Name())
+		buffer.WriteString(";")
+	}
+
+	return buffer.String()
+}
+
+func NewConcurrentTask(name string, tasks ...Task) *ConcurrentTask {
+	return &ConcurrentTask{
+		name:  name,
+		tasks: tasks,
+	}
+}
--- a/tasks/concurrent.go
+++ b/tasks/concurrent.go
@@ -12,6 +12,13 @@ type ConcurrentTask struct {
 	tasks []Task
 }
 
+func NewConcurrentTask(name string, tasks ...Task) *ConcurrentTask {
+	return &ConcurrentTask{
+		name:  name,
+		tasks: tasks,
+	}
+}
+
 func (c *ConcurrentTask) Run() error {
 	var g errgroup.Group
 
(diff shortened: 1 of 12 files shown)


tasks/concurrent.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for _, t := range c.tasks {
+		buffer.WriteString(t.Name())
+		buffer.WriteString(";")
+	}
+
+	return buffer.String()
+}
+
+func NewConcurrentTask(name string, tasks ...Task) *ConcurrentTask {
```
xh3b4sd: The constructors have been on the top so far.

fgimenez: Sure, done.

tasks/concurrent.go
```diff
@@ -0,0 +1,51 @@
...
+	"bytes"
+
+	"github.com/giantswarm/microerror"
+	"golang.org/x/sync/errgroup"
+)
+
+type ConcurrentTask struct {
+	name  string
+	tasks []Task
+}
```
kopiczko: I don't like it groups tasks. It can produce unexpected behavior with other wrappers, like Retry, etc. Also this makes task names awkward.

What do you think about passing `errgroup.Group` to ConcurrentTask instead and wait for the group outside?

fgimenez: > I don't like it groups tasks. It can produce unexpected behavior with other wrappers, like Retry, etc

Could you please explain the problem with retries? AIUI the wrapped tasks are are tasks at the end, they have a run method that blocks and returns an error, we don't care about what happens inside.

> What do you think about passing errgroup.Group to ConcurrentTask instead and wait for the group outside?

I don't think that would work, at the end `ConcurrentTask` is also a `Task`, indistinguishable from any other task, with a `Run` method that blocks and returns an error as above. In this case its `Run` method triggers the concurrent tasks, but from the outside we don't care, we run all the tasks as part of a workflow and all of them are the same https://github.com/giantswarm/architect/blob/master/tasks/spec.go#L19. 

IMO if we make the `ConcurrentTask` different it would break this way of working for no big reason, IMO the only downside is what you mentioned about the naming, but that could be improved, WDYT?

fgimenez:
Build time of aws-operator is down to ~8:45 with a binary built from this branch https://circleci.com/gh/giantswarm/aws-operator/4364, about 70sec less than using architect built from master https://circleci.com/workflow-run/8d0fb1ea-0730-46a2-8c46-335138bd346c

fgimenez:
Also, the pull task is removed from the concurrent set and is executed before it, if not each of the concurrent tasks tries to pull the golang image.

tasks/concurrent.go
kopiczko: When you wrap ConcurrentTask with RetryTask, than all tasks under ConcurrentTask will retry when single fails.

Just to be clear. I think it is difficult to come up with good design now. All approaches have drawbacks. It isn't a blocker for me.

fgimenez: Sure, sorry I didn't get right your concern, what I was asking above was about including a wrapped task in ConcurrentTask, as stated all works as usual, but with the all the tasks (including the wrapped one) executed concurrently. 

ConcurrentTask itself can be wrapped in a retry you want to repeat the whole set on failure (including tasks in the set that potentially have finished). I'm not sure when that can be a valid use case, in general, ConcurrentTask is just a container for a set of tasks that run concurrently, and acts externally as a Task itself, so that it can be run sequentially with other tasks in a workflow or wrapped on a retry (if that's what you want). I've added godoc to ConcurrentTask to make this more clear.

Ok, merging now, thx!

PR #209, opened 2018-03-21
workflow: make go builds and tests less verbose

kopiczko:
When something breaks it's difficult to realize what went wrong. Also circle elides long output.

### DIFF
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -202,7 +202,7 @@ func NewGoTestTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				projectInfo.Project,
 			),
 			Image: fmt.Sprintf("%v:%v", projectInfo.GolangImage, projectInfo.GolangVersion),
-			Args:  []string{"go", "test", "-v", "-race", "./..."},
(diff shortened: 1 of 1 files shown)


MarcelMue:
I don't "get" this, I would prefer as much output as possible usually. This way we don't get into a situation where a build fails and we can't exactly see why. 

Can you give an example for this change?

xh3b4sd:
@MarcelMue can you give an example of a failure in which we do not see what is going wrong? AFAIK in case of failing tests verbosity is implicit for the failing test. When things go wrong you have to debug it anyway locally and run the tests or add more logging to the next CI run. So this here is fine with me personally.

kopiczko:
@MarcelMue when test fails then output is shown for this test (like you passed `-v` flag to this failing case).

Example: https://circleci.com/gh/giantswarm/aws-operator/4527
`Your build output is too large to display in the browser. Only the first 400000 characters are displayed.`

Yes I can download the output. But for what?

Also I don't need to see all compiled packages from vendor directory.

My intention here is to save some time when trying to figure out what went wrong while not sacrificing any valuable information.

I hope that convinces you.

PR #210, opened 2018-03-21
WIP: Only test go packages

indyfree:
Hey guys I need a little bit of help here, also early feedback in general is appreciated <3

I added a condition to the goTestTask, that  only runs when there are actual go packages to be tested.
(We need this for integration test for i.e. helm charts, where no other go code is present than the e2e-test, which have the build constraint)

Problem is, my test of this workflow fails, cause the bash command
`go list ./... <2 | wc -l` is executed in the actual testing directory and not on the `afero.Fs` Filesystem. 
Is there any way around this? I am a bit stuck how to deal with this.

### DIFF
--- a/workflow/error.go
+++ b/workflow/error.go
@@ -6,7 +6,7 @@ import (
 
 var noCertEnvVarError = microerror.New("no cert env var")
 
-// IsNoCertEnvVar asserts multipleHelmChartsError.
+// IsNoCertEnvVar asserts noCertEnvVarError
 func IsNoCertEnvVar(err error) bool {
 	return microerror.Cause(err) == noCertEnvVarError
 }
--- a/workflow/error.go
+++ b/workflow/error.go
@@ -11,6 +11,13 @@ func IsNoCertEnvVar(err error) bool {
 	return microerror.Cause(err) == noCertEnvVarError
 }
 
+var noGolangPackagesError = microerror.New("no golang packages found")
+
+// IsNoGolangPackages asserts noGolangPackagesError
+func IsNoGolangPackages(err error) bool {
+	return microerror.Cause(err) == noGolangPackagesError
+}
+
 var decodeCertError = microerror.New("decode cert")
 
 // IsDecodeCert asserts decodeCertError
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -4,7 +4,10 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/spf13/afero"
 
@@ -61,6 +64,26 @@ func goBuildable(fs afero.Fs, directory string) (bool, error) {
 		if filepath.Ext(fileInfo.Name()) == ".go" {
 			return true, nil
 		}
+	}
+
+	return false, nil
+}
+
+func goTestable() (bool, error) {
+	out, err := exec.Command("bash", "-c", "go list ./... | wc -l").Output()
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	numPackages, err := strconv.Atoi(strings.TrimSpace(string(out)))
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+	if numPackages == 0 {
+		return false, microerror.Mask(noGolangPackagesError)
+	}
+	if numPackages > 0 {
+		return true, nil
 	}
 
 	return false, nil
--- a/workflow/workflow_test.go
+++ b/workflow/workflow_test.go
@@ -302,6 +302,28 @@ func TestGetBuildWorkflow(t *testing.T) {
 			},
 			errorMatcher: IsInvalidHelmDirectory,
 		},
+
+		// Test 11 that a project with only golang files that have build contraints
+		// do not trigger a test workflow.
+		{
+			setUp: func(fs afero.Fs) error {
+				if err := fs.Mkdir(filepath.Join(projectInfo.WorkingDirectory, "integration"), 0644); err != nil {
+					return microerror.Mask(err)
+				}
+				if err := afero.WriteFile(fs, filepath.Join(projectInfo.WorkingDirectory, "integration", "integration_test.go"), []byte("// +build k8srequired"), 0644); err != nil {
+					return microerror.Mask(err)
+				}
+
+				return nil
+			},
+			expectedTaskNames: []string{
+				GoPullTaskName,
+				strings.Join([]string{
+					GoFmtTaskName,
+					GoBuildTaskName,
+				}, ";") + ";",
+			},
+		},
 	}
 
 	for i, tc := range tests {
--- a/workflow/workflow.go
+++ b/workflow/workflow.go
@@ -91,11 +91,17 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 			goTasks = append(goTasks, goBuild)
 		}
 
-		goTest, err := NewGoTestTask(fs, projectInfo)
+		isGoTestable, err := goTestable()
 		if err != nil {
 			return nil, microerror.Mask(err)
 		}
-		goTasks = append(goTasks, goTest)
+		if isGoTestable {
+			goTest, err := NewGoTestTask(fs, projectInfo)
+			if err != nil {
+				return nil, microerror.Mask(err)
+			}
+			goTasks = append(goTasks, goTest)
+		}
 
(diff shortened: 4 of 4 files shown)


workflow/golang.go
```diff
@@ -66,6 +69,26 @@ func goBuildable(fs afero.Fs, directory string) (bool, error) {
 	return false, nil
 }
 
+func goTestable() (bool, error) {
+	out, err := exec.Command("bash", "-c", "go list ./... | wc -l").Output()
```
MarcelMue: I think instead of doing `go list ./ ` you will need to use the actual project path we also use in the tasks for our tests. See here: https://github.com/giantswarm/architect/blob/8b31f756495a11dd3c7572ce460fb4eb2209d2d1/workflow/golang.go#L199

So use ```fmt.Sprintf(
				"/go/src/github.com/%v/%v",
				projectInfo.Organisation,
				projectInfo.Project,
),``` instead

workflow/workflow_test.go
```diff
@@ -302,6 +302,28 @@ func TestGetBuildWorkflow(t *testing.T) {
...
+				}
+
+				return nil
+			},
+			expectedTaskNames: []string{
+				GoPullTaskName,
+				strings.Join([]string{
+					GoFmtTaskName,
+					GoBuildTaskName,
+				}, ";") + ";",
```
MarcelMue: Why is this joined so oddly? Am I missing something?

workflow/golang.go
indyfree: Nice thanks! Gonna try that

workflow/golang.go
```diff
@@ -66,6 +69,26 @@ func goBuildable(fs afero.Fs, directory string) (bool, error) {
...
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	numPackages, err := strconv.Atoi(strings.TrimSpace(string(out)))
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+	if numPackages == 0 {
+		return false, microerror.Mask(noGolangPackagesError)
```
MarcelMue: Why are we returning an error here? Shouldn't this be the case where we just return false because we have nothing to test?

workflow/golang.go
```diff
@@ -66,6 +69,26 @@ func goBuildable(fs afero.Fs, directory string) (bool, error) {
...
+	}
+	if numPackages > 0 {
+		return true, nil
+	}
+
+	return false, nil
```
MarcelMue: Would this ever be reached? `numPackages == 0` and `numPackages > 0` should cover all possible outcomes.

workflow/golang.go
indyfree: Yep your right, thats actually completely wrong.

workflow/golang.go
indyfree: Good catch. Adressed that by only assesing the positive case and otherwise return false

workflow/workflow_test.go
indyfree: Appearently thats the way architect bundles the `go-fmt`, `go-build`, and `go-test` tasks together. 
``` bash
2018/03/20 16:52:14 running workflow: {
	retry task 'go-pull:	'docker pull quay.io/giantswarm/golang:1.10.0''
	go-fmt;go-build;go-test;
}
```

Just went with the flow there, see 
https://github.com/giantswarm/architect/blob/d87bc24ffef75d9ba84f5be3c8300d1b9e35911a/workflow/workflow_test.go#L109-L113
https://github.com/giantswarm/architect/blob/d87bc24ffef75d9ba84f5be3c8300d1b9e35911a/workflow/workflow_test.go#L132-L135

MarcelMue: Okay, fine by me :+1:

fgimenez: Yes, this is how a `ConcurrentTask` `String` method output looks like https://github.com/giantswarm/architect/blob/master/tasks/concurrent.go#L48 suggestions to improve are more than welcome :)

workflow/golang.go
fgimenez: Why not `return numPackages > 0, nil`?

workflow/error.go
```diff
@@ -6,7 +6,7 @@ import (
 
 var noCertEnvVarError = microerror.New("no cert env var")
 
-// IsNoCertEnvVar asserts multipleHelmChartsError.
+// IsNoCertEnvVar asserts noCertEnvVarError
```
fgimenez: <3

workflow/golang.go
fgimenez: Also IMO no need for bash and the pipe, just calling `go list` in the right dir and counting the output lines in stdout should be enough.

indyfree: The problem remains that the tests and the other conditions e.g. `goBuildable()` make use of `afero.Fs` as an filesystem abstraction to test if files exists. But a command like `go list` doesn't see the the test files created by `fs` in the workflow test, but works on a real filesystem. Maybe @fgimenez has an idea?

fgimenez: Not sure, maybe you can use a combination of `Cmd.StdinPipe` https://golang.org/src/os/exec/exec.go?s=14969:15018#L521 and `afero.WriteReader` https://github.com/spf13/afero/blob/master/util.go#L40 to connect the standard input of the command to the afero `io.Writer` passing it a precreated directory (in afero fs) with different go files and packages per test case. Would be very interesting to see that working :)

workflow/golang.go
```diff
@@ -66,6 +68,20 @@ func goBuildable(fs afero.Fs, directory string) (bool, error) {
 	return false, nil
 }
 
+func goTestable(directory string) (bool, error) {
+	cmd := exec.Command("go", "list", "./...")
+	cmd.Dir = directory
+
+	out, err := cmd.Output()
```
indyfree: Still stuck here. The implementation is the right one IMO. But build-workflow tests fail because the `go list` command is executed on the OS filesytem. The filesystem used for testing is the mocked `afero MemMapFs`. I don't see any way to connect the `afero fs` as input for `go list` and also not sure if thats the right way to change the implementation because the tests lack.

Another approach would be to use `afero.NewOsFs()` instead of `afero.NewMemMapFs()` for the build-workflow test to execute the tests on the OSs filesystem. It's usually not recommended to use the `OsFs` for tests, because it would need more path configuration and cleanup steps, but I don't have any other idea. @fgimenez WDYT?

fgimenez: Sorry about the problem with `go list` and `exec.Command`, we can pipe the list of files to the standard input of the command (and this list can come from `afero.MemMapFs`), but it needs anyway to access them to figure out the actual list of packages, and for that it uses plain `os` methods.

IMO as we talked before that's better than walking the directories and searching for build tags, :+1: to creating a tempdir with `OsFs` for each testcase.

indyfree: Cool thanks. Just to get it right because the formulation is a bit unclear to me:

I am gonna switch to `OsFs` to create tempdirs for the BuildWorkflow right?

indyfree: Yes, was thinking of something like this:
```
		fs := afero.NewOsFs()

		// Create testing dir and clean up after
		fs.Mkdir(testdir, 0666)
		defer fs.RemoveAll(testdir)
```

fgimenez: I would make sure that each test case uses a different test dir with `testdir, err := afero.TempDir(fs,"", "test-dir")`, the code above can't be run in parallel for instance.

indyfree: Oh yes thanks, didn't think of parallelism

workflow/workflow_test.go
```diff
@@ -290,23 +298,54 @@ func TestGetBuildWorkflow(t *testing.T) {
...
+		// Test 11 that a project with only golang files that have build contraints
+		// do not trigger a test workflow.
+		{
+			setUp: func(fs afero.Fs, testDir string) error {
+				projectInfo.WorkingDirectory = testDir
+
+				if err := fs.Mkdir(filepath.Join(projectInfo.WorkingDirectory, "integration"), 0744); err != nil {
+					return microerror.Mask(err)
+				}
+				if err := afero.WriteFile(fs, filepath.Join(projectInfo.WorkingDirectory, "integration", "integration_test.go"), []byte("// +build k8srequired\n\npackage test"), 0644); err != nil {
```
indyfree: Do we usually wrap lines? Wrapped is looks :hankey: too ;)

MarcelMue: I think this is mostly up to you, I don't have a big preference either way.

PR #212, opened 2018-03-27
Fix publish command

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/1902.

Make that publish command actually work.

### DIFF
--- a/cmd/publish.go
+++ b/cmd/publish.go
@@ -25,7 +25,7 @@ var (
 )
 
 func init() {
-	RootCmd.AddCommand(deployCmd)
+	RootCmd.AddCommand(publishCmd)
 
 	var defaultDockerUsername string
 	var defaultDockerPassword string


kopiczko:
What does publish actually do? Was it unused so far?

(I added sig-releng to reviewers)

indyfree:
Yes publish was unused so far. We introduced it to be able to push/publish charts to different channels on quay.io. That is needed for the release workflow of guest cluster charts of the managed services story. See:
https://github.com/giantswarm/giantswarm/blob/master/roadmap/MANAGED-SERVICES.md#releases--channels
https://github.com/giantswarm/giantswarm/blob/master/roadmap/MANAGED-SERVICES.md#release-promotion
PR: https://github.com/giantswarm/architect/pull/190

indyfree:
merged #212.

PR #218, opened 2018-04-16
Add unpublish command

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/1902, https://github.com/giantswarm/giantswarm/issues/2706.

Add `architect unpublish` to cleanup channels on the CNR after pushing to temporary channels. This needed e.g. in guest cluster charts, where channels get created only for a specific CI helm test run.

### DIFF
--- a/workflow/helm.go
+++ b/workflow/helm.go
@@ -162,7 +162,7 @@ func NewHelmDeleteFromChannelTask(fs afero.Fs, chartDir string, projectInfo Proj
 			Args: []string{
 				"registry",
 				"channel",
-				fmt.Sprintf("%v/%v/%v/%s", projectInfo.Registry, projectInfo.Organisation, projectInfo.Project, "-chart"),
+				fmt.Sprintf("%v/%v/%v%s", projectInfo.Registry, projectInfo.Organisation, projectInfo.Project, "-chart"),
 				"--delete",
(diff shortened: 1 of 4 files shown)


workflow/helm.go
```diff
@@ -142,3 +143,30 @@ func NewHelmPromoteToChannelTask(fs afero.Fs, chartDir string, projectInfo Proje
...
+		tasks.DockerTaskConfig{
+			WorkingDirectory: chartDir,
+			Image:            HelmImage,
+			Volumes: []string{
+				fmt.Sprintf("%v:/root/.cnr/", cnrDir),
+			},
+			Args: []string{
+				"registry",
+				"channel",
+				fmt.Sprintf("%v/%v/%v%s", projectInfo.Registry, projectInfo.Organisation, projectInfo.Project, "-chart"),
```
xh3b4sd: We should also start using the appr and helm client libs here.

workflow/workflow_test.go
```diff
@@ -602,3 +602,96 @@ func TestGetPublishWorkflow(t *testing.T) {
 		})
 	}
 }
+
+func TestGetUnpublishWorkflow(t *testing.T) {
```
fgimenez: Very nice!

PR #236, opened 2018-07-03
unify and simplify project list

calvix:
towards https://github.com/giantswarm/giantswarm/issues/3473

this diff is super weird, sorry
basically I just moved the list into 3 vars for aws, azure and kvm and then they are reused ( and in one case we append extra component)


calvix:
@JosephSalisbury WDYT ?

events/events.go
```diff
@@ -108,256 +23,56 @@ var environmentProjects = map[Environment][]string{
...
-		"companyd",
-		"credentiald",
-		"desmotes",
-		"draughtsman",
-		"etcd-backup",
-		"g8s-cert-manager",
+	}
+	// aws project list + aws specific services
+	awsProjectList = append(baseProjectList,
 		"g8s-cloudwatch-exporter",
```
JosephSalisbury: The OCD side of me would alphabetise this

events/events.go
```diff
@@ -108,256 +23,56 @@ var environmentProjects = map[Environment][]string{
...
+)
+
+// Environment is a name of an installation.
+type Environment string
+
+// environmentProjects is a mapping between Environments,
+// and the projects that should be deployed there.
+// We use codenames for installations to not expose customer information.
+var environmentProjects = map[Environment][]string{
+	"centaur": append(kvmProjectList, "pv-cleaner-operator"),
```
JosephSalisbury: Do we want to list these alphabetically too? Just to make it easier to find a specific installations.

calvix: oh Jesus the ordernacis

PR #253, opened 2018-09-27
remove-route53-from-girafe

calvix:
temporary solution for https://github.com/giantswarm/giantswarm/issues/4282


events/events.go
```diff
@@ -81,7 +81,7 @@ var environmentProjects = map[Environment][]string{
 	"geckon":    kvmProjectList,
 	"ghost":     azureProjectList,
 	"ginger":    awsProjectList,
-	"giraffe":   awsChinaProjectList,
+	"giraffe":   awsProjectList,
```
calvix: only difference between these two is the `route53-manager`

PR #262, opened 2018-11-23
tasks: always remove docker containers

kopiczko:
Not sure what the issue was about. The URL in the comment gives 404.
I found this snippet (from
https://circleci.com/docs/2.0/building-docker-images/):

	- run: |
	    # start container with the application
	        # make sure you're not using `--rm` option otherwise the
	        container will be killed after finish
	            docker run --name app app-image:1.2.3

It looks like they are now fine with using --rm.

### DIFF
--- a/tasks/docker.go
+++ b/tasks/docker.go
@@ -1,9 +1,5 @@
 package tasks
 
-import (
-	"os"
-)
-
 type DockerTaskConfig struct {
 	Volumes          []string
 	Env              []string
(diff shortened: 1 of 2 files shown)


kopiczko:
> Just FMI what was the issue?

It wasn't the issue. This was just bugging me after porting dockerized go builds to `e2e-harnes`.

kopiczko:
I'm confident after https://github.com/giantswarm/e2e-harness/pull/170 works correctly.

PR #268, opened 2019-01-11
Fail builds if DCO or LICENSE files are missing

JosephSalisbury:
Closes https://github.com/giantswarm/giantswarm/issues/5029

This will apply to *all* repos that `architect` builds, but I think having a switch based on public/private isn't necessary - if we should do it for public repos, we should do it for private repos too.

### DIFF
--- a/DCO
+++ b/DCO
@@ -1,0 +1,36 @@
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
+660 York Street, Suite 102,
+San Francisco, CA 94110 USA
+
+Everyone is permitted to copy and distribute verbatim copies of this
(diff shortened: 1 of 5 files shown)


JosephSalisbury:
I won't merge this till Monday at the earliest, and I'll make sure to give a good shout in #news-dev first.

kopiczko:
What should we do with private repo? Put Apache2 license?

JosephSalisbury:
Yeah!

marians:
Why also check private repos?

JosephSalisbury:
Mainly to keep it simpler / have one less rule to remember.

JosephSalisbury:
https://media.giphy.com/media/g0mNCFSdDSjscSc5q9/giphy.gif

PR #269, opened 2019-01-14
Remove ingress-operator from projects list

r7vme:
related: https://github.com/giantswarm/giantswarm/issues/3952


calvix:
we could even have SRE task to remove them from installations

PR #273, opened 2019-01-31
Remove license text validation

rossf7:
See https://github.com/giantswarm/credentiald/pull/27#pullrequestreview-198669036 and https://gigantic.slack.com/archives/C3C7ZQXC1/p1548933728017700.

The extra license text validation is causing some confusion in PRs and not really helping. 

Licensing is a complex topic (because lawyers  ;) so this PR removes the validation. We can re-enable when its clear what approach we want to take.

### DIFF
--- a/workflow/repo.go
+++ b/workflow/repo.go
@@ -2,8 +2,6 @@ package workflow
 
 import (
 	"fmt"
-	"strings"
-	"time"
 
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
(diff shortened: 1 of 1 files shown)


kopiczko:
I don't understand why we should drop it. This is legit to make sure we have this text in the license. I'd even go a step further and make the SHA check against the LICENSE. That way we're can be sure its identical everywhere to the single byte :)

rossf7:
Closing as there have been no more requests to change the current logic.

PR #274, opened 2019-02-14
set git commit SHA in pkg/project.gitSHA

TheoBrigitte:
Towrads: https://github.com/giantswarm/giantswarm/issues/5371
Follow up: https://github.com/giantswarm/release-operator/pull/50

We introduce the new `pkg/project` package in release-operator, which regroup project wide information like git sha, project name, etc.

Architect now needs to update the git sha, in the `pkg/project` package instead of `main`.

### DIFF
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -258,8 +258,10 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				"-a",
 				"-v",
 				"-ldflags", fmt.Sprintf(
-					"-w -X main.gitCommit=%s -linkmode 'auto' -extldflags '-static'",
+					"-w -X main.gitCommit=%s -X github.com/%s/%s/pkg/project.gitSHA=%[1]s -linkmode 'auto' -extldflags '-static'",
 					projectInfo.Sha,
+					projectInfo.Organisation,
+					projectInfo.Project,
 				),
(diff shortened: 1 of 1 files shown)


workflow/golang.go
```diff
@@ -258,8 +258,10 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				"-a",
 				"-v",
 				"-ldflags", fmt.Sprintf(
-					"-w -X main.gitCommit=%s -linkmode 'auto' -extldflags '-static'",
+					"-w -X main.gitCommit=%s -X github.com/%s/%s/pkg/project.gitSHA=%[1]s -linkmode 'auto' -extldflags '-static'",
```
xh3b4sd: I would add the sha again instead of doing the `%[1]s` magic.

workflow/golang.go
```diff
@@ -258,8 +258,10 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
...
 					projectInfo.Sha,
+					projectInfo.Organisation,
+					projectInfo.Project,
```
kopiczko: `source` and `version` are still not injected.

workflow/golang.go
kopiczko: +1

PR #277, opened 2019-02-22
Add a way to create deployment events for only a specific group of testing environments.

oponder:
This should allow me to do: `architect deploy --group=testing` in CI, to conditionally deploy to only our testing environments.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -103,12 +103,42 @@ var environmentProjects = map[Environment][]string{
 	},
 }
 
+// testingGroup is a specific grouping of environments that are considered
+// testing environments.
+var testingGroup = []Environment{
+	"gauss",
+	"geckon",
+	"ghost",
+	"ginger",
+	"giraffe",
+	"godsmack",
+	"gorgoth",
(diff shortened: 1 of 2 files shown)


events/events.go
```diff
@@ -103,12 +103,42 @@ var environmentProjects = map[Environment][]string{
 	},
 }
 
+// testingGroup is a specific grouping of environments that are considered
+// testing environments.
+var testingGroup = []Environment{
```
teemow: Hardcoding this is not a good idea imo. We already have the information about the pipeline in the control planes.

oponder: hmm so you're thinking of emit deployment events for all environments every time but then with some kind of metadata so that draughtsman can do the filtering?

teemow: I guess there is no need to design this properly. The hardcoded projects and environments are already in here. I don't care much. We just need to get rid of this in architect as a whole at some point.

oponder: @teemow if you're ok with it, can you approve?

PR #279, opened 2019-02-25
detect git tags

TheoBrigitte:

This PR add enables architect to detect git tags.


cmd/root.go
```diff
@@ -57,6 +58,17 @@ func init() {
 		defaultSha = strings.TrimSpace(string(out))
 	}
 
+	// Use git tag as defaultRef when available otherwise use defaultSha.
+	var defaultRef string
+	{
+		out, err := exec.Command("git", "describe", "--tags", "--exact-match", "HEAD").Output()
+		if err != nil {
+			defaultRef = defaultSha
```
TheoBrigitte: `defaultSha` is used here so later on we can make direct use of `ref` as docker image tag for example without additional conditions.

cmd/root.go
```diff
@@ -57,6 +58,17 @@ func init() {
 		defaultSha = strings.TrimSpace(string(out))
 	}
 
+	// Use git tag as defaultRef when available otherwise use defaultSha.
+	var defaultRef string
```
kopiczko: For my interest - why is it called `ref`?

PR #280, opened 2019-02-26
detect git tag

TheoBrigitte:

This PR enables architect to detect git tags.


cmd/root.go
```diff
@@ -57,6 +58,17 @@ func init() {
 		defaultSha = strings.TrimSpace(string(out))
 	}
 
+	// Use git tag as defaultRef when available otherwise use defaultSha.
+	var defaultRef string
+	{
+		out, err := exec.Command("git", "describe", "--tags", "--exact-match", "HEAD").Output()
+		if err != nil {
+			defaultRef = defaultSha
```
TheoBrigitte: `defaultSha` is used here so later on we can make direct use of `ref` as docker image tag for example without additional conditions.

cmd/root.go
```diff
@@ -74,6 +86,7 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&project, "project", defaultProject, "name of the project")
 
 	RootCmd.PersistentFlags().StringVar(&branch, "branch", defaultBranch, "git branch to build")
+	RootCmd.PersistentFlags().StringVar(&sha, "ref", defaultRef, "git ref to build")
```
kopiczko: Should we have this flag? Shouldn't this be always inferred from the SHA? Setting this to a wrong value may cause inconsistencies.

kopiczko:
```suggestion
	RootCmd.PersistentFlags().StringVar(&ref, "ref", defaultRef, "git ref to build")
```

cmd/root.go
```diff
@@ -57,6 +58,17 @@ func init() {
 		defaultSha = strings.TrimSpace(string(out))
 	}
 
+	// Use git tag as defaultRef when available otherwise use defaultSha.
+	var defaultRef string
```
TheoBrigitte: > For my interest - why is it called ref?

@kopiczko It is called ref, because it can hold either a git commit sha or a git tag. It is arguable that technically a git sha is not a reference but I haven't come up with a better name so far. Any suggestion are welcome.

cmd/root.go
TheoBrigitte: Yes that's a valid point, I'll remove the argument.

cmd/root.go
kopiczko: Ack. Makes sense. Now when we don't have a flag, shouldn't we just set the `ref`? I was trying to figure why we have `defaultRef` variable.

TheoBrigitte: Removed `defaultRef`

PR #281, opened 2019-02-26
update docker workflow

TheoBrigitte:
Towards: [giantswarm/giantswarm#5206](https://github.com/giantswarm/giantswarm/issues/5206)

- ~update `emptyShaError` to `emptyRefError`~
- ~rename `sha` to `ref` in different places.~
- add git tag to `projectInfo`
- tag and push docker image with git tag

PR #282, opened 2019-02-26
New DockerTag and Version argument for helm charts

TheoBrigitte:

* add `DockerTag` as a new argument for helm charts
* add `Version` as a new argument for helm charts

This will makes the transition smoother for repositories which needs to support git tag, as it enforce to change the helm chart to make use of the new `DockerTag` argument for the docker image tag.


template/template.go
```diff
@@ -13,6 +13,8 @@ var (
 
 // BuildInfo holds information concerning the current build.
 type BuildInfo struct {
+	// Ref is the reference of the commit we are building for.
+	Ref string
```
kopiczko: Given my comment in #281 we should push the image for both SHA and tag I think it would make sense to rename that to `Tag` and keep it empty when there is none. Then architect should do things only when the tag is non-empty.

TheoBrigitte: Good point, let's no couple things.

TheoBrigitte:
The idea is to be able to template chart with the correct information.

I am aiming at having chart in the following format
Actually we cannot have an empty value here, because this is what we use for templating the helm chart.


```Chart.yaml
# Chart.yaml
name: hello-test-chart
version: [[ .Version ]]
```
```templates/deployment.yaml
# templates/deployment.yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: hello-test
spec:
  template:
    spec:
      containers:
      - name: hello-test
        image: quay.io/giantswarm/hello-test:[[ .DockerTag ]]
```

Where :
- `Version` will be either a git tag or a git tag +git suffix.
- `DockerTag` will be either a git commit sha or a git tag

TheoBrigitte:
ping @giantswarm/team-magic

template/helm_test.go
```diff
@@ -13,15 +13,19 @@ import (
...
+		version   string
+		setUp     func(afero.Fs, string) error
+		check     func(afero.Fs, string) error
 	}{
 		// Test that a chart is templated correctly.
 		{
-			chartDir: "/helm/test-chart",
-			sha:      "jabberwocky",
+			chartDir:  "/helm/test-chart",
+			dockerTag: "white-rabbit",
```
JosephSalisbury: <3

TheoBrigitte: Feel to good to have my work appreciated <3

PR #283, opened 2019-02-26
New ReleaseGithubTask to create github releases

TheoBrigitte:

Provide a new `ReleaseGithubTask` which allow to create draft github releases and with assets files.

Note: for this to work we have to provide a github token with full repo access, see https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes


release/release.go
```diff
@@ -0,0 +1,86 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/google/go-github/github"
+
+	"github.com/giantswarm/microerror"
+)
+
+// CreateWithDir creates a draft github release with files contained in dir as assets.
+func CreateWithDir(client *github.Client, dir, organisation, project, sha, tag string) error {
```
kopiczko: Why do we need to expose that method? I would suggest making it private method of `ReleaseGithubTask` to avoid passing some depedencies around.

kopiczko: Number of string args here is concerning me. I would replace that with input struct to avoid potential bugs.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+func create(client *github.Client, organisation, project, sha, tag string, draft bool) (*github.RepositoryRelease, error) {
+	release := &github.RepositoryRelease{
+		Draft:           &draft,
+		Name:            &tag,
+		TagName:         &tag,
+		TargetCommitish: &sha,
+	}
+
+	createdRelease, _, err := client.Repositories.CreateRelease(
+		context.TODO(),
```
kopiczko: Please pass the context to this function. If there is need to TODO let's do it in single place.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+		organisation,
+		project,
+		release,
+	)
+	if err != nil {
+		return nil, microerror.Maskf(err, "could not create release for %s:%s", project, tag)
```
kopiczko: We're already working to remove possibility of masking non microerror.Error errors. Let's not annotate this one.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+	return createdRelease, nil
+}
+
+func uploadAsset(client *github.Client, organisation, project string, id int64, file *os.File) error {
+	options := &github.UploadOptions{
+		Name: filepath.Base(file.Name()),
+	}
+
+	_, _, err := client.Repositories.UploadReleaseAsset(
+		context.TODO(),
```
kopiczko: Ditto TODO context.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+		organisation,
+		project,
+		id,
+		options,
+		file,
+	)
+	if err != nil {
+		return microerror.Maskf(err, "could not upload release asset id:%d file:%s", id, file.Name())
```
kopiczko: Ditto error masking.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// CreateWithDir creates a draft github release with files contained in dir as assets.
+func CreateWithDir(client *github.Client, dir, organisation, project, sha, tag string) error {
+	release, err := create(client, organisation, project, sha, tag, false)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	filesInfo, err := ioutil.ReadDir(dir)
```
kopiczko: Is `ReadDir` recursive? I think it isn't.

release/release.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	for _, f := range filesInfo {
```
kopiczko: Shouldn't we create a tarball instead of uploading all the files?

release/task.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+const (
+	ReleaseGithubTaskName   = "release-github"
+	ReleaseGithubTaskString = "%s: sha:%s tag:%s"
+)
+
+type ReleaseGithubTask struct {
+	client       *github.Client
+	dir          string
```
kopiczko: What is the dir here? Isn't it better to pass the common `ProjectInfo` struct?

release/release.go
kopiczko: I think this function should be idempotent. What if we try to recreate a release when the build is reran?

kopiczko:
> Note: for this to work we have to provide a github token with full repo access, see https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes

Did we have read-only so far?

release/release.go
TheoBrigitte: No `ioutil.ReadDir` is not recursive. It just list the content of the given directory.

release/release.go
TheoBrigitte: This task upload whatever is found in the directory, the helm chart tarball will be created by a different task [`PackageHelmChartTask`](https://github.com/giantswarm/architect/pull/284)

TheoBrigitte:
> > Note: for this to work we have to provide a github token with full repo access, see https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes
> 
> Did we have read-only so far?

So far we had `repo_deployment` which allow us to create github deployment events.

workflow/github.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	if projectInfo.Project == "" {
+		return microerror.Mask(emptyProjectError)
+	}
+	if projectInfo.Ref == "" {
+		return microerror.Mask(emptyRefError)
+	}
+	if projectInfo.Sha == "" {
+		return microerror.Mask(emptyShaError)
+	}
```
kopiczko: I don't like we have so many defined errors that are unhandled, but I understand it's a copy/paste.

release/release.go
kopiczko: It's a bit confusing but I don't have a better idea.

release/release.go
TheoBrigitte: fixed: this is now private

release/release.go
TheoBrigitte: fixed: added context

release/release.go
TheoBrigitte: fixed: removed error annotation

release/release.go
TheoBrigitte: This will probably make more sense in later PR when all of this will be connected.

release/task.go
TheoBrigitte: reworked: dir is now called assetDir and `ProjectInfo` struct is also used

release/release.go
TheoBrigitte: reworked this: `ensureWithDir` is now idempotent

TheoBrigitte: fixed: added `releaseInfo` struct

PR #284, opened 2019-02-26
New PackageHelmChartTask to package helm charts

TheoBrigitte:

Provide a new `PackageHelmChartTask` which allow to package helm chart like `helm package` does.


pack/pack.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+
+	// Load chart from a directory.
+	ch, err := chartutil.LoadDir(path)
+	if err != nil {
+		return err
+	}
+
+	// Save the chart as an archive in the given director.
```
kopiczko: I would just skip those comments.

```suggestion
	// Save the chart as an archive in the given directory.
```

PR #286, opened 2019-03-01
add version

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5371

From [voo spec](https://github.com/giantswarm/giantswarm/blob/master/specs/versioning-of-operators.md#ci):
> - Version tag is injected into the "version" variable in package "./pkg/project" during build time.
>     - When the tag is not set for the commit the variable is set to the latest tag in the branch with "+git" suffix. E.g. "v1.2.0+git".

This PR adds version detection to architect.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -81,7 +81,7 @@ func init() {
 	{
 		// version can be of three different formats:
 		//   v1.0.0: building a tagged version.
-		//   v1.0.0+git: building ahead of a tagged version.
+		//   v1.0.0-3a955cbb126f0fe5d51aedf2eb84acca7b074374: building ahead of a tagged version.
 		//   v0.0.0-939f5c6949f83c0a7ea98a25bc9524fd2f751ffe: building a repo which has no tags.
 		if tag != "" {
 			version = tag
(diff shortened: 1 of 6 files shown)


cmd/root.go
```diff
@@ -66,14 +69,32 @@ func init() {
...
+			version = strings.TrimSpace(string(out))
+		}
+	}
+
+	// Define the version we are building.
+	{
+		// version can be of three different formats:
+		//   v1.0.0: building a tagged version.
+		//   v1.0.0+git: building ahead of a tagged version.
+		//   v0.0.0-939f5c6949f83c0a7ea98a25bc9524fd2f751ffe: building a repo which has no tags.
```
TheoBrigitte: I didn't knew what to do in case there is no tag in the repository, so I went with the go modules approach of making up a fake v0.0.0-<sha> version. This lead me to the idea of also using the `-<sha>` suffix instead of `+git` in case we are building ahead of a tag. So we would get a more precise information baked in the version.

JosephSalisbury: Building a version for a repo without tags would be pretty rare, right? Being as we have to create a tag for something to get into a release. I'm fine with the fake v0.0.0 thing.

And having a proper sha if we're ahead of a tag makes a lot of sense. I like!

kopiczko: I thing following go modules approach makes total sense 👍 Let's also update the spec.

PR #287, opened 2019-03-05
Inject version into go binary

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/5371

Inject the version being built into the go binary at `pkg/project.version`
Also provide test for `NewGoBuildTask`.

### DIFF
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -261,9 +261,12 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 					"-w "+
 						"-linkmode 'auto' "+
 						"-extldflags '-static' "+
-						"-X 'main.gitCommit=%[1]s' "+
-						"-X 'github.com/%s/%s/pkg/project.gitSHA=%[1]s' "+
+						"-X 'main.gitCommit=%s' "+
+						"-X 'github.com/%s/%s/pkg/project.gitSHA=%s' "+
 						"-X 'github.com/%s/%s/pkg/project.version=%s'",
+					projectInfo.Sha,
+					projectInfo.Organisation,
+					projectInfo.Project,
 					projectInfo.Sha,
 					projectInfo.Organisation,
 					projectInfo.Project,
--- a/workflow/golang.go
+++ b/workflow/golang.go
@@ -257,21 +257,14 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				"go", "build",
 				"-a",
 				"-v",
-				"-ldflags", fmt.Sprintf(
-					"-w "+
-						"-linkmode 'auto' "+
-						"-extldflags '-static' "+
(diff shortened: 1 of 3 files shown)


workflow/golang.go
```diff
@@ -258,10 +258,16 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				"-a",
 				"-v",
 				"-ldflags", fmt.Sprintf(
-					"-w -X main.gitCommit=%s -X github.com/%s/%s/pkg/project.gitSHA=%[1]s -linkmode 'auto' -extldflags '-static'",
+					"-w "+
+						"-linkmode 'auto' "+
+						"-extldflags '-static' "+
+						"-X 'main.gitCommit=%[1]s' "+
+						"-X 'github.com/%s/%s/pkg/project.gitSHA=%[1]s' "+
+						"-X 'github.com/%s/%s/pkg/project.version=%s'",
```
TheoBrigitte: TIL: 

> After processing a bracketed expression [n], subsequent verbs will use arguments n+1, n+2, etc. unless otherwise directed.
see https://golang.org/pkg/fmt/

It feels a bit like black magic, any suggestions to make this looks better appreciated.

xh3b4sd: Well yeah, don't use that shit. Provide the information as you need them, even if you have to repeat the input a little bit. Another alternative would be a proper template.

TheoBrigitte: removed the wizardry thingy

workflow/golang.go
```diff
@@ -258,10 +258,19 @@ func NewGoBuildTask(fs afero.Fs, projectInfo ProjectInfo) (tasks.Task, error) {
 				"-a",
 				"-v",
 				"-ldflags", fmt.Sprintf(
-					"-w -X main.gitCommit=%s -X github.com/%s/%s/pkg/project.gitSHA=%[1]s -linkmode 'auto' -extldflags '-static'",
+					"-w "+
+						"-linkmode 'auto' "+
+						"-extldflags '-static' "+
+						"-X 'main.gitCommit=%s' "+
+						"-X 'github.com/%s/%s/pkg/project.gitSHA=%s' "+
+						"-X 'github.com/%s/%s/pkg/project.version=%s'",
```
kopiczko: How about

```go
"-ldflags", strings.Join([]string{
	"-w",
	"-linkmode 'auto'",
	"-extldflags '-static'",
	fmt.Sprintf("-X 'main.gitCommit=%s'", projectInfo.Sha),
	fmt.Sprintf("-X 'github.com/%s/%s/pkg/project.gitSHA=%s'", projectInfo.Organisation, projectInfo.Project, projectInfo.Sha),
	fmt.Sprintf("-X 'github.com/%s/%s/pkg/project.version=%s'", projectInfo.Organisation, projectInfo.Project, projectInfo.Version),
}, " "),
```

?

TheoBrigitte: I like it it looks clean

PR #288, opened 2019-03-08
add test for PackageHelmChartTask

TheoBrigitte:

Add unit test for helm package task (`PackageHelmChartTask`).

1. create a chart
2. package the chart
3. check that the created helm chart archive contains what we expect

### DIFF
--- a/pack/pack_test.go
+++ b/pack/pack_test.go
@@ -1,0 +1,178 @@
+package pack
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/giantswarm/microerror"
+	"k8s.io/helm/pkg/chartutil"
+	hapichart "k8s.io/helm/pkg/proto/hapi/chart"
+)
+
+var (
+	chartYaml = `name: hello-test-chart
+version: v1.0.0`
+	valuesYaml     = `namespace: giantswarm`
+	deploymentYaml = `apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: hello-test
+  namespace: {{ .Values.namespace }}
+  labels:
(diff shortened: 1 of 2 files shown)


pack/pack_test.go
```diff
@@ -0,0 +1,178 @@
...
+		expectedChartName      string
+		expectedChartVersion   string
+		expectedDeploymentName string
+		expectedDeploymentData string
+		expectedValues         string
+
+		setup func(chartDir string) error
+	}{
+		{
+			name:                   "case 0: test package chart",
```
MarcelMue: Would be nice to have a second test to check what happens if one of the inputs is missing. e.g. no `Chart.yaml`. Or no `values.yaml`.

TheoBrigitte:
> LGTM. Does it already have multi-chart support?

This has no support for multi-chart, as multi-chart is built in the [`processHelmDir`](https://github.com/giantswarm/architect/blob/master/workflow/workflow.go#L315) function. I will link those together in a later PR.

pack/pack_test.go
TheoBrigitte: added 2 more test:
* no chart, which yield `"chart metadata (Chart.yaml) missing"` error
* only Chart.yaml, which is the minimum valid chart.

pack/pack_test.go
```diff
@@ -0,0 +1,228 @@
...
+				}
+
+				return createFiles(files)
+			},
+		},
+		{
+			name: "case 1: no chart",
+			errorMatcher: func(err error) bool {
+				expectedError := errors.New("chart metadata (Chart.yaml) missing")
+				return err.Error() == expectedError.Error()
```
TheoBrigitte: I'am not proud of this, but that's how the error is defined in helm: https://github.com/helm/helm/blob/master/pkg/chartutil/load.go#L200

PR #291, opened 2019-03-16
add release command

TheoBrigitte:

Provide `architect release` command, whenever a git tag is present, it creates a draft github release, with helm charts uploaded as assets.

### DIFF
--- a/cmd/release.go
+++ b/cmd/release.go
@@ -1,0 +1,87 @@
+package cmd
+
+import (
+	"context"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/google/go-github/github"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"golang.org/x/oauth2"
+
+	"github.com/giantswarm/architect/tasks"
+	"github.com/giantswarm/architect/workflow"
+)
+
+var (
+	releaseCmd = &cobra.Command{
+		Use:   "release",
+		Short: "release chart as github release",
+		Run:   runRelease,
+	}
+)
+
+func init() {
+	RootCmd.AddCommand(releaseCmd)
+}
+
+func runRelease(cmd *cobra.Command, args []string) {
+	var err error
+
+	ctx := context.Background()
+
+	fs := afero.NewOsFs()
+
+	projectInfo := workflow.ProjectInfo{
+		WorkingDirectory: workingDirectory,
+		Organisation:     organisation,
+		Project:          project,
+
+		Sha: sha,
+		Tag: tag,
+
+		Version: version,
+	}
+
+	var githubClient *github.Client
+	{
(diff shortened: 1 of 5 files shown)


cmd/release.go
```diff
@@ -0,0 +1,87 @@
...
+		Run:   runRelease,
+	}
+)
+
+func init() {
+	RootCmd.AddCommand(releaseCmd)
+}
+
+func runRelease(cmd *cobra.Command, args []string) {
+	var err error
```
xh3b4sd: Can we please have something like calling `runReleaseError` here and have proper error handling like anywhere else plus proper logging configured? From a software engineering point of view it would be really beneficial holding up our high standards and not writing random code. We do not have to reinvent the wheel and maintaining such different code will only be more complicated down the road. You may also want to read https://github.com/giantswarm/giantswarm/blob/master/sigs/operator/command_line_tools.md. We could also think about extending the docs with more pitfalls we find here.

kopiczko: I'd also like to have program termination handled in a single place (or as few as possible).

cmd/release.go
```diff
@@ -0,0 +1,87 @@
...
+		if err != nil {
+			log.Fatalf("could not get workflow: %v", err)
+		}
+
+		log.Printf("running workflow: %s\n", workflow)
+
+		if dryRun {
+			log.Printf("dry run, not actually running workflow")
+			return
+		}
```
kopiczko: This dry-run feels useless, but I guess it's useless anyway.

TheoBrigitte:
Restructured the release command:
* moved into `cmd/release` package
* using cobra RunE in order to exit program in a single place
* added errors definition

cmd/root.go
```diff
@@ -80,32 +83,44 @@ func init() {
...
 	RootCmd.PersistentFlags().StringVar(&organisation, "organisation", defaultOrganisation, "organisation who owns the project")
 	RootCmd.PersistentFlags().StringVar(&project, "project", defaultProject, "name of the project")
 
-	RootCmd.PersistentFlags().StringVar(&branch, "branch", defaultBranch, "git branch to build")
-	RootCmd.PersistentFlags().StringVar(&sha, "sha", defaultSha, "git SHA1 to build")
+	RootCmd.PersistentFlags().StringVar(&branch, "branch", defaultBranch, "git branch being built")
+	RootCmd.PersistentFlags().StringVar(&sha, "sha", defaultSha, "git SHA1 being built")
+	RootCmd.PersistentFlags().String("tag", defaultTag, "git tag being built")
+
+	RootCmd.PersistentFlags().String("version", defaultVersion, "project version being built")
```
TheoBrigitte: Setting all values as flag is required in order to be able to access them from subcommands using `Command.Flags()`.

TheoBrigitte:
Need a second review round, please

PR #293, opened 2019-04-02
support legacy version, trim v prefix

TheoBrigitte:

In order to not introduce breaking changes, architect need to be able to continue building charts using the old `1.0.0-<sha>` version. Otherwise draughtman will break see https://github.com/giantswarm/azure-operator/pull/499#issuecomment-479013968.

Also remove the `v` prefix from the version being detected as helm charts do not support this. Additionally docker images tag are preferred to be without this prefix.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -68,7 +68,7 @@ func init() {
 	{
 		out, err := exec.Command("git", "describe", "--tags", "--exact-match", "HEAD").Output()
 		if err == nil {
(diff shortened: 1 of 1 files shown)


kopiczko:
So we have docker tags and helm charts without `v` and git tags with `v` correct?

TheoBrigitte:
> So we have docker tags and helm charts without `v` and git tags with `v` correct?

Yes that's correct.

PR #294, opened 2019-04-02
forgot the cmd/release/flag.go file

TheoBrigitte:

Forgot to add `cmd/release/flag.go` in previous PR.

### DIFF
--- a/cmd/release/flag.go
+++ b/cmd/release/flag.go
@@ -1,0 +1,11 @@
+package release
+
+import "os"
+
+func init() {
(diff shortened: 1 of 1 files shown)


TheoBrigitte:
In this case login info are used to log into quay.io helm registry, so falling back to `~/.docker/config.json` would not work here I think.

https://github.com/giantswarm/architect/blob/a4aa62329b1a977133ce28b8e49ee591ec9bbd97/workflow/helm.go#L73-L88 
It would be good to refactor this code, since we are mixing docker and helm login info here.

PR #296, opened 2019-04-03
only build tag on circleci when explicitly requested

TheoBrigitte:
Towards: https://github.com/giantswarm/azure-operator/pull/499

Fixes an issue where `architect` is building tagged version even if CircleCI is requesting to build regular commit sha.

See https://circleci.com/gh/giantswarm/azure-operator/5229

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -68,7 +68,10 @@ func init() {
 	{
 		out, err := exec.Command("git", "describe", "--tags", "--exact-match", "HEAD").Output()
 		if err == nil {
-			defaultTag = strings.TrimPrefix(strings.TrimSpace(string(out)), "v")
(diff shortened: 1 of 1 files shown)


tuommaki:
Just for clarification: So when Circle is building a branch (e.g. for PR), with this change it doesn't assign tag linker parameter to compiler when building unless explicitly requested (which then occurs only with tagging, and master builds, I guess)?

TheoBrigitte:
Yes, the corner case here is the following:
1. create a new commit
2. tag this commit
3. push both tag and commit `git push --tags origin HEAD`
4. CircleCI launch two builds (1 for the commit, 1 for the tag)

The actual issue here is that we detect the tag using `git describe` in architect which in this case would yield a tag for both builds, and this is wrong in the case of the commit build, there we want to build using the sha and not the tag.

Issue #157, opened 2017-09-01
architect build should have a switch for pushing images


JosephSalisbury:

webwurst:
Would this solve the following?
```
architect build
```
```
2017/10/18 16:27:58 could not fetch workflow: empty docker username
```

oponder:
I need this flag so I can enable forked pull requests on Happa.

I want to build but not push for forked pull requests, since pushes will fail without our secrets in the ENV.

Happy to take a try at the implementation.

@JosephSalisbury how does something like `architect build --no-push` sound to you?

JosephSalisbury:
Alternatively, we can be a bit clever and not try to push if the credentials aren't present (as I assume they wouldn't be for forked pull requests).

PR #300, opened 2019-05-08
Add a --no-push flag that skips the docker push on 'architect build'

oponder:
This is kind of the dumbest thing I could think of.

Alternatively / Ideally, this would be split into two workflows and commands? so we'd have `architect build` and `architect push` ?

That does mean editing a lot of circleci files in our other repos though. So I thought I'd start with this to see what you guys think.

The resulting double negative in the code is not ideal. (if !noPush) i.e.: If not not push. :P 

Towards: https://github.com/giantswarm/architect/issues/157


cmd/build.go
```diff
@@ -55,6 +57,8 @@ func init() {
 
 	buildCmd.Flags().StringVar(&golangImage, "golang-image", "quay.io/giantswarm/golang", "golang image")
 	buildCmd.Flags().StringVar(&golangVersion, "golang-version", "1.11.3", "golang version")
+
+	buildCmd.Flags().BoolVar(&noPush, "no-push", false, "don't push the built image to the registry")
```
MarcelMue: Why not have the flag called `push` and default to `true`? Makes the whole negation thing easier.

oponder: yeah that makes sense 👍

oponder: but wait... how does that look in the command line?

how do you 'unset' a flag that is default set to true?

oponder: --push=false ? is that possible? I'll see

MarcelMue: Yes that is how it should work AFAIK.

oponder: alright done!

oponder:
@JosephSalisbury @MarcelMue PTAL / PTALA ?

TheoBrigitte:
I like the approach of being able to build the code but not push it. I think that we should also prevent the helm chart to be pushed, WDYT ?
https://github.com/giantswarm/architect/blob/d1f4cb037d3d73ce3be4fc568ae666af012ed58d/workflow/workflow.go#L185


In such case the docker image would be built and available locally only on the machine it was built, so would be the helm chart. No one else could actually make use of the helm chart since the docker image referenced in it would not be available.

oponder:
@TheoBrigitte Good point. I didn't notice the helm part of this workflow. Will include it in the conditional.

workflow/workflow.go
```diff
@@ -145,49 +145,51 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
...
+				}
 			}
 		}
-	}
 
-	helmTasks, err := processHelmDir(fs, projectInfo, NewHelmPushTask)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+		helmTasks, err := processHelmDir(fs, projectInfo, NewHelmPushTask)
```
TheoBrigitte: I would say that we still want the helm chart to be templated locally, but not pushed.

oponder: ok I think i get what you mean, and I implemented a Noop function.. It feels a bit slimy.. 

so just want to see if this is what you had in mind, I can clean it up, possibly give it its own definition instead of defining it inline and add some comments explaining the intention.

TheoBrigitte: Yeah that's what I had in mind, but I'm not so confident this will run smoothly since we basically return a nil `Task` here, a, have you tested it ?

If that run great, otherwise you will have to implement a noop task, so return a real [`Task`](https://godoc.org/github.com/giantswarm/architect/tasks#Task) which does nothing.

oponder: Yeah I did a `architect build` and it went ok. But I'll try doing it in a real CI build with happa later.

I'm also possibly doing some timeboxing to see if I can solve this in a way that feels a bit cleaner.

TheoBrigitte: all good then, code looks correct to me.

TheoBrigitte:
Just gave this a try, it ran but threw some errors, here is the ouput
```
$ architect build --push=false
2019/05/18 17:10:11 running workflow: %!s(PANIC=String method: runtime error: invalid memory address or nil pointer dereference)
2019/05/18 17:10:11 running task: repo-check
repo has required file 'DCO'
repo has required file 'LICENSE'
repo 'LICENSE' has required text 'Copyright 2016 - 2019 Giant Swarm GmbH'
2019/05/18 17:10:11 running task: retry task 'go-pull:  'docker pull quay.io/giantswarm/golang:1.11.3''
1.11.3: Pulling from giantswarm/golang
54f7e8ac135a: Already exists
d6341e30912f: Already exists
087a57faf949: Already exists
5d71636fb824: Already exists
f368dba6a331: Already exists
69ffff7c9ff9: Already exists
2613ffabce6c: Already exists
Digest: sha256:b85fdf27f598d39826447e7e516d5afc4225954f7e49de43955831298b3d98f6
Status: Image is up to date for quay.io/giantswarm/golang:1.11.3
2019/05/18 17:10:14 running task: go-fmt;go-build;go-test;
< lines removed >
2019/05/18 17:30:52 generating Junit report at: /var/folders/g0/z1wf7ry91lbcct5ghd7vwvgm0000gn/T/results/golang/results.xml
2019/05/18 17:30:52 could not execute workflow: <nil>
```

The two following lines were produced by the changes here and the lack of a proper task
`2019/05/18 17:10:11 running workflow: %!s(PANIC=String method: runtime error: invalid memory address or nil pointer dereference)`
`2019/05/18 17:30:52 could not execute workflow: <nil>`

TheoBrigitte:
I added a noop task to fix the issue and here is the new output
```
$ architect build --push=false  --docker-username theo@giantswarm.io --docker-password xxx
2019/05/18 17:35:39 running workflow: {
        repo-check
        retry task 'go-pull:    'docker pull quay.io/giantswarm/golang:1.11.3''
        go-fmt;go-build;go-test;
        docker-build:   'docker build -t quay.io/giantswarm/aws-operator:7eb5f09e525f196f90ebe69cc9f329152197e3f5 /Users/theo/src/github.com/giantswarm/aws-operator'
        docker-run-version:     'docker run --rm -w /Users/theo/src/github.com/giantswarm/aws-operator quay.io/giantswarm/aws-operator:7eb5f09e525f196f90ebe69cc9f329152197e3f5 version'
        docker-run-help:        'docker run --rm -w /Users/theo/src/github.com/giantswarm/aws-operator quay.io/giantswarm/aws-operator:7eb5f09e525f196f90ebe69cc9f329152197e3f5 --help'
        retry task 'helm-pull:  'docker pull quay.io/giantswarm/docker-helm:23ac1ca6527bb59e07a6a8c36809488e39dd5791''
        helm-login:     'docker run --rm -v /Users/theo/.cnr:/root/.cnr/ -w  quay.io/giantswarm/docker-helm:23ac1ca6527bb59e07a6a8c36809488e39dd5791 registry login --user=theo@giantswarm.io --password=[REDACTED] quay.io'
        template-helm-chart:    /Users/theo/src/github.com/giantswarm/aws-operator/helm/aws-operator-chart docker-tag:7eb5f09e525f196f90ebe69cc9f329152197e3f5 sha:7eb5f09e525f196f90ebe69cc9f329152197e3f5 version:8-7eb5f09e525f196f90ebe69cc9f329152197e3f5
        retry task 'noop task'
}
2019/05/18 17:35:39 running task: repo-check
< lines removed >
2019/05/18 17:45:51 running task: retry task 'noop task'
2019/05/18 17:45:51 generating Junit report at: /var/folders/g0/z1wf7ry91lbcct5ghd7vwvgm0000gn/T/results/golang/results.xml
```

oponder:
Thx theo! Nice that you drove it home.

I'll give this a try in a Happa forked branch workflow now

oponder:
Dang, almost there.

Still need to get around the helm login task, since that will fail in the forked pr environment as well.

Line 377 in workflow.go `helmLogin, err := NewHelmLoginTask(fs, projectInfo)`

`processHelmDir` seems to do a bit too much

TheoBrigitte:
> processHelmDir seems to do a bit too much

I agree on this. IMO the best way would be to move the helm login out of `processHelmDir`, and implements it where its needed.
Otherwise we can pass the push flag to `processHelmDir` but this sounds bit hacky again.

oponder:
Closing, I probably won't get around to making the requested changes.

PR #303, opened 2019-05-15
Adding circleci-exporter

tomahawk28:
- Adding `circleci-exporter` into the `gollum` installation.


marians:
@tomahawk28 Out of curiosity, why is this done? circle-ci exporter is not deployed on cluster `6iec4` on gollum.

PR #306, opened 2019-05-19
Add push and destination flag to release command

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6025

With the new push-to-catalog approach, we need architect to be able to package helm chart locally but prevent it to publish them as github releases (in case of test versions).

This PR introduces 2 new flags:
- `--push`: when set to false this prevent the helm chart to be published as github release.
- `--destination`: destination folder where to put the helm chart archives.
It also fixes the logic for `destination/releaseDir` directory which was quite broken.

### DIFF
--- a/cmd/release/flag.go
+++ b/cmd/release/flag.go
@@ -13,4 +13,5 @@ func init() {
 	Cmd.Flags().String("docker-username", defaultDockerUsername, "username to use to login to docker registry")
 	Cmd.Flags().String("docker-password", defaultDockerPassword, "password to use to login to docker registry")
 	Cmd.Flags().BoolVar(&push, "push", true, "publish helm chart as a github release")
+	Cmd.Flags().String("destination", "architect-release", "destination folder")
 }
--- a/cmd/release/flag.go
+++ b/cmd/release/flag.go
@@ -2,10 +2,15 @@ package release
 
 import "os"
 
+var (
+	push bool
+)
+
 func init() {
 	defaultDockerUsername := os.Getenv("QUAY_USERNAME")
 	defaultDockerPassword := os.Getenv("QUAY_PASSWORD")
 
 	Cmd.Flags().String("docker-username", defaultDockerUsername, "username to use to login to docker registry")
 	Cmd.Flags().String("docker-password", defaultDockerPassword, "password to use to login to docker registry")
+	Cmd.Flags().BoolVar(&push, "push", true, "publish helm chart as a github release")
 }
--- a/cmd/release/runner.go
+++ b/cmd/release/runner.go
@@ -2,9 +2,9 @@ package release
 
 import (
 	"context"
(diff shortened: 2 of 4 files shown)


cmd/release/flag.go
```diff
@@ -2,10 +2,16 @@ package release
...
+	push bool
+)
+
 func init() {
 	defaultDockerUsername := os.Getenv("QUAY_USERNAME")
 	defaultDockerPassword := os.Getenv("QUAY_PASSWORD")
 
 	Cmd.Flags().String("docker-username", defaultDockerUsername, "username to use to login to docker registry")
 	Cmd.Flags().String("docker-password", defaultDockerPassword, "password to use to login to docker registry")
+	Cmd.Flags().BoolVar(&push, "push", true, "publish helm chart as a github release")
```
MarcelMue: Why is the flag called push and not publish?

cmd/release/flag.go
```diff
@@ -2,10 +2,16 @@ package release
...
+	Cmd.Flags().String("destination", "architect-release", "destination folder")
```
MarcelMue: There should be some info here that it only applies if push is false, right?

cmd/release/flag.go
TheoBrigitte: I wanted to follow the same direction as in https://github.com/giantswarm/architect/pull/300. But publish makes also sense, no big opinion on this.

cmd/release/flag.go
TheoBrigitte: No, this is always used in every cases.

cmd/release/flag.go
MarcelMue: There is this confusing naming going on: Helm talks about publish in its CLI IIRC  for pushing charts while docker uses push. I am ultimately fine with either but would like some more opinions on this in general.

TheoBrigitte: I would go with `push`, we will have this soon for architect release command, plus this align with docker and git.

cmd/release/runner.go
```diff
@@ -49,17 +49,27 @@ func runReleaseError(cmd *cobra.Command, args []string) error {
...
+		path := filepath.Join(cmd.Flag("working-directory").Value.String(), cmd.Flag("destination").Value.String())
+		_, err := os.Stat(path)
 		if err != nil {
-			return microerror.Mask(err)
+			if os.IsNotExist(err) {
+				err := os.Mkdir(path, os.ModePerm)
+				if err != nil {
+					return microerror.Mask(err)
+				}
+			} else {
```
kopiczko: Why not `} else if !os.IsExist(err) {`?

TheoBrigitte:
Closing this as we will make use of `helm package` instead of architect to package charts.

PR #310, opened 2019-05-31
container for architect

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6025

Create a docker image for architect to be used from circleci.
This docker image contains architect + other tools we need.

Tools required from our side:
* architect
* git
* helm

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM alpine:3.9
+
+ARG HELM_VERSION=v2.14.0
+
+RUN apk add --no-cache \
+        ca-certificates \
+        curl \
+        git \
+        openssh-client &&\
+        # Install helm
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM alpine:3.9
+
+ARG HELM_VERSION=v2.14.0
+
+RUN apk add --no-cache \
+        ca-certificates \
```
JosephSalisbury: Is it really nitpicky to ask towards having specific versions pinned in the Dockerfile?

TheoBrigitte: I think this is legit.

JosephSalisbury: Okay cool, then let’s do that :D I’m bored of dependencies changing underneath us

TheoBrigitte: done. PTALA

PR #322, opened 2019-07-11
Fix error handling in helm templating walk

sslavic:
This PR fixes error handling in helm templating walk. It also handles case when walk is called with non-nil error passed in.

Bug was uncovered by @pipo02mix when trying to `architect helm template` https://github.com/giantswarm/prometheus-operator-app/tree/master/helm/kubernetes-prometheus-chart 
See related discussion on slack: https://gigantic.slack.com/archives/C3C7ZQXC1/p1562842254031900


kopiczko:
Please request review for sig-releng for architect.

PR #323, opened 2019-07-11
Revert "Fix error handling in helm templating walk (#322)"

sslavic:
This reverts commit e8a2207a087142b2ebdc00723e281f40d09083d8.


rossf7:
I approved to get CI up and running again. But I'd like to understand why we can't return errors in this context 😕

PR #324, opened 2019-07-11
Make helm template tgz check nil safe

sslavic:
This PR makes `architect helm template` tgz check nil safe.

I've tested it on https://github.com/giantswarm/prometheus-operator-app/tree/master/helm/kubernetes-prometheus-chart - before PR `architect helm template` would segfault, and after it completes successfully.

```
$ architect helm template --dir ./helm/prometheus-operator-app-chart
2019/07/12 00:03:17 templating helm chart
dir: ./helm/prometheus-operator-app-chart
sha: e877bd37b17d9b7e75d762ec61ac2a3d73220962
tag: 0.1.0
version: 0.1.0
2019/07/12 00:03:17 templated helm chart
```

### DIFF
--- a/template/helm.go
+++ b/template/helm.go
@@ -41,7 +41,7 @@ type TemplateHelmChartTask struct {
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 func (t TemplateHelmChartTask) Run() error {
 	err := afero.Walk(t.fs, t.chartDir, func(path string, info os.FileInfo, err error) error {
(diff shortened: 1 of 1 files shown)


xh3b4sd:
Can we then also fix the error handling as you did in your other PR?

sslavic:
@xh3b4sd I had to revert that error handling fix, it broke many builds.
With this PR I'd like to unblock @pipo02mix as soon as possible and this was most straight forward fix for the problem he was facing, without changing rest of the walk function behavior.

PR #327, opened 2019-07-25
remove support for .DockerTag in helm template

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6557

This is too confusing to have `.DockerTag` and `.Version` when templating helm charts, so we decided to only have `.Version` which we use to tag the docker image and the chart. So everything get the same version tag.
Simplicity is beauty 🌟 

⚠️ Do not merge before `.DockerTag` is removed from aws-operator and azure-operator helm charts.

### DIFF
--- a/template/helm.go
+++ b/template/helm.go
@@ -31,7 +31,8 @@ const (
 )
 
 type TemplateHelmChartTask struct {
-	fs       afero.Fs
+	fs afero.Fs
+
 	chartDir string
 	sha      string
 	version  string
--- a/template/helm.go
+++ b/template/helm.go
@@ -17,8 +17,8 @@ const (
 
 	// TemplateHelmChartTaskString is the format for printing the
 	// helm chart templating task.
-	// Name of the task, the helm directory path, the docker-tag, the sha, and the version.
-	TemplateHelmChartTaskString = "%s:\t%s docker-tag:%s sha:%s version:%s"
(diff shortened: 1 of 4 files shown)


template/helm.go
```diff
@@ -31,11 +31,10 @@ const (
 )
 
 type TemplateHelmChartTask struct {
-	dockerTag string
-	fs        afero.Fs
-	chartDir  string
-	sha       string
-	version   string
+	fs       afero.Fs
```
rossf7: Nitpicky but I think we should separate the dependences and settings with a newline here.

template/template.go
```diff
@@ -13,8 +13,6 @@ var (
 
 // BuildInfo holds information concerning the current build.
 type BuildInfo struct {
-	// DockerTag is the docker image tag.
-	DockerTag string
```
kopiczko: Don't we need to fix charts first?

I think at least aws-operator and release-operator are still using those.

TheoBrigitte: aws-operator and azure-operator are no longer using this.

https://github.com/giantswarm/aws-operator/pull/1804
https://github.com/giantswarm/azure-operator/pull/548

template/helm.go
TheoBrigitte: Better ?

PR #328, opened 2019-07-29
template: add replace template function

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/6557

See https://github.com/giantswarm/aws-operator/pull/1804/files#diff-39f31ed2a4f9740e98c9adb74ad80904R12.

### DIFF
--- a/template/helm.go
+++ b/template/helm.go
@@ -56,13 +56,28 @@ func (t TemplateHelmChartTask) Run() error {
 			Version:   t.version,
 		}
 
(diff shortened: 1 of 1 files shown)


template/helm.go
```diff
@@ -56,13 +56,28 @@ func (t TemplateHelmChartTask) Run() error {
 			Version:   t.version,
 		}
 
-		newTemplate := template.Must(template.New(path).Delims("[[", "]]").Parse(string(contents)))
-		if err != nil {
```
kopiczko: This also fixes the error handling. `Must` would cause panic and this err is from elsewhere.

PR #332, opened 2019-09-05
new command: create appcr

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/6847

This PR implements a new command to create app cr.

```yaml
$ architect create appcr --name test --app-name foo --app-version bar --catalog baz
apiVersion: application.giantswarm.io/v1alpha1
kind: App
metadata:
  creationTimestamp: null
  labels:
    app-operator.giantswarm.io/version: 1.0.0
  name: test
  namespace: giantswarm
spec:
  catalog: baz
  kubeConfig:
    inCluster: true
  name: foo
  namespace: giantswarm
  version: bar
status: {}
```

This PR replaces: https://github.com/giantswarm/opsctl/pull/412


cmd/create/appcr/flag.go
```diff
@@ -0,0 +1,9 @@
+package appcr
+
+func init() {
```
MarcelMue: I know we do this here but I hate it with a passion.

cmd/create/command.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/architect/cmd/create/appcr"
+)
+
+var (
+	Cmd = &cobra.Command{
+		Use:   "create",
+		Short: "create all the things.",
```
MarcelMue: Something better here would be nice.

pkg/app/app.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return appCR
+}
+
+func Print(w io.Writer, format string, appCR *applicationv1alpha1.App) error {
+	var output []byte
+	var err error
+
+	switch format {
+	case "json":
```
kopiczko: Why do we need json output?

pkg/app/app.go
```diff
@@ -0,0 +1,77 @@
+package app
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+
+	"github.com/ghodss/yaml"
```
kopiczko: Didn't you want to use this magical k8s library because of empty fields?

pkg/app/error.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/giantswarm/microerror"
+)
+
+var wrongFormatError = &microerror.Error{
+	Kind: "wrongFormatError",
+}
+
+func IsWrongFormatError(err error) bool {
```
kopiczko: Not handled so it should be replaced with `executionFailedError`.

pkg/app/app.go
```diff
@@ -0,0 +1,77 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Namespace is the namespace where App CRs are created.
+	namespace = "giantswarm"
+
+	labelAppOperatorVersion = "app-operator.giantswarm.io/version"
+	labelReleaseCyclePhase  = "release-operator.giantswarm.io/release-cycle-phase"
+	labelServiceType        = "giantswarm.io/service-type"
```
kopiczko: I don't see labelReleaseCyclePhase and labelServiceType being used.

pkg/app/app.go
```diff
@@ -0,0 +1,77 @@
...
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Namespace is the namespace where App CRs are created.
+	namespace = "giantswarm"
+
+	labelAppOperatorVersion = "app-operator.giantswarm.io/version"
```
kopiczko: Can we use string literal instead of the const? What is the added value here?

cmd/create/command.go
TheoBrigitte: Going point, that was my first intention then I forgot

pkg/app/app.go
TheoBrigitte: We do not but in case it's there.

pkg/app/app.go
TheoBrigitte: Yes, it does the job correctly while the k8s `PrintObj` gave me a hard time with dependencies.

kopiczko: Nice 👍

pkg/app/app.go
TheoBrigitte: removed those

pkg/app/app.go
TheoBrigitte: Using string literal now, I originally copied the whole thing from release-operator,

pkg/app/error.go
TheoBrigitte: done

cmd/create/appcr/flag.go
TheoBrigitte: I hate it too, so I changed it.

PR #333, opened 2019-09-13
dockerfile: install go 1.13.0

kopiczko:
- Install go 1.13.0
- Update to alpine 3.10
- Use rettaged images
- Keep curl
- Remove versions from installed binaries


kopiczko:
Going with this as I need it for the new job in architect-orb I'm crafting right now.

PR #338, opened 2019-10-11
Go modules

TheoBrigitte:
Move to go modules.


.circleci/config.yml
```diff
@@ -1,21 +1,25 @@
-version: 2
-jobs:
-  build:
-    machine: true
-    steps:
-    - checkout
+version: 2.1
```
sslavic: @kopiczko for 2.1 one has to "Enable pipelines" in CircleCI project advanced settings. Without this project was in "Needs setup" state, didn't build the commit which switched to architect-orb, had a misleading error that Workflows require version 2. I just enabled pipelines for the CircleCI architect project and requested Github CircleCI to resend notification about architect-orb commit. Build triggered this time and execution failed (see [here](https://circleci.com/gh/giantswarm/architect/961)). Now it complains about missing `deploy` job definition - it seems it can't be defined inline in workflow definition.

kopiczko: Thanks. I'll fix it.

TheoBrigitte:
Looks fine, can I get approval here?

go.mod
```diff
@@ -0,0 +1,26 @@
+module github.com/giantswarm/architect
+
+go 1.13
```
rossf7: In some places go version is at the top. Others at the bottom. Nit, I'd like to have this aligned.

rossf7: Damn too slow! 😉 

Anyway not a blocker.

TheoBrigitte: `go mod tidy` is generating this and we have CI check in place for this.

kopiczko: I guess it makes sense to have module at the top?

rossf7: I find it easier to scan the file in devctl. But what I mainly care about is consistency across projects.

https://github.com/giantswarm/devctl/blob/f39ef486eb7dd167213e2e705e0c25539a4ef08a/go.mod

kopiczko: I see. I would propose having `module` line first. `go` line second and `require` third. WDYT?

rossf7: Yes I agree that is the best order. 👍

PR #341, opened 2019-10-15
circleci: don't call architect deploy on release

kopiczko:


xh3b4sd:
Why though?

MarcelMue:
> Why though?

The orb does it further down in the file already.

kopiczko:
To be precise. The orb jobs do everything `architect deploy` would do. Also it doesn't work work because VOO/legacy versioning is incompatible.

PR #344, opened 2019-11-15
hook: use gitrepo

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/7491

```
$ cd aws-operator

$ git rev-parse origin/HEAD
fa8ef815f789dd6bcb9680a3402f7696eba93b95

$ architect project version
7.0.0-fa8ef815f789dd6bcb9680a3402f7696eba93b95
```


cmd/hook/preRunE.go
```diff
@@ -1,66 +1,57 @@
 package hook
```
xh3b4sd: Why is the file name partially upper case? We are not doing this usually. Any strong reason for that?

cmd/hook/preRunE.go
```diff
@@ -1,66 +1,57 @@
...
+	var err error
+
+	var repo *gitrepo.Repo
 	{
-		out, err := exec.Command("git", "rev-parse", "HEAD").Output()
+
+		dir, err := gitrepo.TopLevel(ctx, ".")
 		if err != nil {
-			return microerror.Maskf(gitNoSHAError, "could not get git sha: %#v\n", err)
+			return microerror.Mask(err)
```
xh3b4sd: <3

cmd/hook/preRunE.go
kopiczko: I didn't even realize. No, I think it's wrong.

PR #350, opened 2019-11-21
add azure-app-collection and kvm-app-collection

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/7720

Adding the new [azure-app-collection](https://github.com/giantswarm/azure-app-collection) and [kvm-app-collection](https://github.com/giantswarm/kvm-app-collection) to the list of project.


events/events.go
```diff
@@ -45,13 +45,15 @@ var (
...
 		"azure-operator",
+		"azure-app-collection",
 	)
 	// kvm project list + kvm specific services
 	kvmProjectList = append(baseProjectList,
 		"bridge-operator",
 		"flannel-operator",
 		"ingress-exporter",
 		"kvm-operator",
+		"kvm-app-collection",
```
kopiczko: Won't it error when those repos don't exist or it's unrelated?

TheoBrigitte: Yes this would fail when repos don't exist, but they do exist now.

PR #351, opened 2019-11-22
helm-chart-template: template only the values.yaml and Chart.yaml files

piontec:
Fixes giantswarm/giantswarm#7614

### DIFF
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -22,18 +22,23 @@ func runTemplateError(cmd *cobra.Command, args []string) error {
 
 	log.Printf("templating helm chart\ndir: %s\nsha: %s\ntag: %s\nversion: %s\n", chartDir, sha, tag, version)
 
-	helmChartTemplate, err := helmtemplate.NewTemplateHelmChartTask(helmtemplate.Config{
-		Fs:       fs,
-		ChartDir: chartDir,
-		Sha:      sha,
-		Version:  version,
-	})
+	var err error
+	var s *helmtemplate.TemplateHelmChartTask
+	{
+		c := helmtemplate.Config{
+			Fs:       fs,
+			ChartDir: chartDir,
+			Sha:      sha,
+			Version:  version,
+		}
 
-	if err != nil {
-		return microerror.Mask(err)
+		s, err = helmtemplate.NewTemplateHelmChartTask(c)
+		if err != nil {
+			return microerror.Mask(err)
+		}
 	}
 
-	if err := helmChartTemplate.Run(); err != nil {
+	if err := s.Run(); err != nil {
 		return microerror.Mask(err)
 	}
 
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -7,7 +7,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
-	"github.com/giantswarm/architect/workflow"
+	"github.com/giantswarm/architect/helmtemplate"
 )
 
 func runTemplateError(cmd *cobra.Command, args []string) error {
@@ -18,17 +18,17 @@ func runTemplateError(cmd *cobra.Command, args []string) error {
 		version  = cmd.Flag("version").Value.String()
 	)
 
-	var projectInfo = workflow.ProjectInfo{
-		Sha:     sha,
-		Tag:     tag,
-		Version: version,
-	}
-
 	fs := afero.NewOsFs()
 
 	log.Printf("templating helm chart\ndir: %s\nsha: %s\ntag: %s\nversion: %s\n", chartDir, sha, tag, version)
 
-	helmChartTemplate, err := workflow.NewTemplateHelmChartTask(fs, chartDir, projectInfo)
+	helmChartTemplate, err := helmtemplate.NewTemplateHelmChartTask(helmtemplate.Config{
+		Fs:       fs,
+		ChartDir: chartDir,
+		Sha:      sha,
+		Version:  version,
+	})
+
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/helmtemplate/helm.go
+++ b/helmtemplate/helm.go
@@ -1,0 +1,115 @@
+package helmtemplate
+
+import (
+	"bytes"
+	"fmt"
+	"path"
+	"text/template"
+
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/afero"
+)
+
+const (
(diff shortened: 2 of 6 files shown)


piontec:
> That would be too easy. We need to only change `architect helm template` command used by architect orb. Other functionality must remain untouched. Otherwise we'll have to deal with too much backward compatibility.

Wait, what else this PR is changing?

kopiczko:
`architect deploy` for example. Anything that uses `workflow` package (which uses `template` package).

piontec:
> `architect deploy` for example. Anything that uses `workflow` package (which uses `template` package).

You're right, for some reason VSCode hasn't shown me anything outside the `cmd` and `template` packages as for where this is used.

kopiczko:
I use `grep` for that - it never fails :D

piontec:
Still, this is a blocker for me and I need to fix it. Unfortunately, I know nothing about other architect's use cases, so the only other solution I can come up with is add a flag to the templating call: either do `SimplifidTemplating` or `FullTemplating`. Make `Full` backward compatible default and pass `Simple` only from `cmd/helm/template`. Will this work?

kopiczko:
I'd be fine with providing another package used exclusively by `architect helm template` as the rest is obsolete and going to be dropped when we switch to orb fully.

piontec:
This works for me, but I'm really not sure if that's what you mean. Especially, I'm not sure about the `TemplateHelmChartTaskName` var.

template/error.go
```diff
@@ -4,47 +4,53 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
-var incorrectShaError = &microerror.Error{
+// IncorrectShaError shows that SHA is wrong
+var IncorrectShaError = &microerror.Error{
```
kopiczko: Why are errors exported now?

template/template.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 var (
-	// permission is a default permission to use for templated files
-	permission os.FileMode = 0644
+	// Permission is a default Permission to use for templated files
+	Permission os.FileMode = 0644
```
kopiczko: Just copy it over to the new package.

helmtemplate/helm.go
```diff
@@ -0,0 +1,92 @@
...
+func (t TemplateHelmChartTask) Name() string {
+	return TemplateHelmChartTaskName
+}
+
+func (t TemplateHelmChartTask) String() string {
+	return fmt.Sprintf(TemplateHelmChartTaskString, t.Name(), t.chartDir, t.sha, t.version)
+}
+
+// NewTemplateHelmChartTask creates a new TemplateHelmChartTask
+func NewTemplateHelmChartTask(fs afero.Fs, chartDir, sha, version string) TemplateHelmChartTask {
```
kopiczko: When we are constructing a new package it would be good to align with the latest conventions. See https://github.com/giantswarm/fmt/blob/master/go/struct_initialization.md.

helmtemplate/helm.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"bytes"
+	"fmt"
+	"path"
+	"text/template"
+
+	"github.com/spf13/afero"
+
+	archtemp "github.com/giantswarm/architect/template"
+	"github.com/giantswarm/microerror"
```
kopiczko: `github.com/giantswarm/microerror` should be moved up. https://github.com/giantswarm/fmt/blob/master/go/imports.md

piontec:
OK, PTAL again

helmtemplate/helm.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if config.Sha == "" {
+		return nil, microerror.Maskf(incorrectValueError, "%T.Sha must not be empty", config)
+	}
+
+	if config.Version == "" {
+		return nil, microerror.Maskf(incorrectValueError, "%T.Version must not be empty", config)
+	}
+
+	return &TemplateHelmChartTask{
```
kopiczko: We usually assign this to the same name as in methods selector:

```suggestion
	t := &TemplateHelmChartTask{
```

helmtemplate/helm.go
```diff
@@ -0,0 +1,115 @@
...
+		fs:       config.Fs,
+		chartDir: config.ChartDir,
+		sha:      config.Sha,
+		version:  config.Version,
+	}, nil
```
kopiczko: And then:

```suggestion
	}
	
	return t, nil
```

Again https://github.com/giantswarm/fmt/blob/master/go/struct_initialization.md:

```
	s := &Struct{
		logger:     config.Logger,

		fullName: fmt.Sprintf("%s-%d", config.Name, config.Order),

		name:    config.Name,
		timeout: timeout,
	}

	return s, nil
```

cmd/helm/template/runner.go
```diff
@@ -18,17 +18,17 @@ func runTemplateError(cmd *cobra.Command, args []string) error {
...
 
 	log.Printf("templating helm chart\ndir: %s\nsha: %s\ntag: %s\nversion: %s\n", chartDir, sha, tag, version)
 
-	helmChartTemplate, err := workflow.NewTemplateHelmChartTask(fs, chartDir, projectInfo)
+	helmChartTemplate, err := helmtemplate.NewTemplateHelmChartTask(helmtemplate.Config{
+		Fs:       fs,
+		ChartDir: chartDir,
+		Sha:      sha,
+		Version:  version,
+	})
```
kopiczko: Let's please stick to the pattern described here https://github.com/giantswarm/fmt/blob/master/go/struct_initialization.md

```
	var s *Struct
	{
		c := Config{
			Logger: logger,

			Name:    "Helicopter",
			Order:   1,
			Timeout: 15,
		}

		s, err = New(c)
		if err != nil {
			return microerror.Mask(err)
		}
	}
```

piontec:
I think I got yours 'just copy over' wrong. I'll update today or on monday.

helmtemplate/helm.go
```diff
@@ -0,0 +1,104 @@
...
+	Fs afero.Fs
+
+	ChartDir string
+	Sha      string
+	Version  string
+}
+
+// Run templates the chart's Chart.yaml and templates/deployment.yaml.
+func (t TemplateHelmChartTask) Run() error {
+
```
kopiczko:
```suggestion
```

piontec:
OK, can you help me with releasing this now? What's the process and how it works? Should I merge it now, assign a tag or something?

kopiczko:
Once you merge this will be used in all chart builds using the orb (https://github.com/giantswarm/architect-orb/). I.e. ones that are landing in chart catalog.

PR #359, opened 2020-02-11
Allow to customize channel where to push when building

fiunchinho:
This solution sucks.

But we are currently pushing to the stable channel while building PR's, which means that unexpected changes from unmerged PR's may end up in our e2e tests which install stable charts from apiextensions.


cmd/build.go
```diff
@@ -58,6 +59,8 @@ func init() {
 
 	buildCmd.Flags().StringVar(&golangImage, "golang-image", "quay.io/giantswarm/golang", "golang image")
 	buildCmd.Flags().StringVar(&golangVersion, "golang-version", "1.13.1", "golang version")
+
+	publishCmd.Flags().StringVar(&channels, "channels", "", "channels to publish the charts to, separated by comma")
```
kopiczko: How would that work? Would we have now to update all the CI configs and set the channel if the branch is `master`?

rossf7:
```suggestion
	publishCmd.Flags().StringVar(&channels, "channels", "stable", "channels to publish the charts to, separated by comma")
```

I think it should default to stable for backwards compatibility. So we only change the circle config if we see this problem.

workflow/helm.go
```diff
@@ -86,6 +87,19 @@ func NewHelmPushTask(fs afero.Fs, chartDir string, projectInfo ProjectInfo) (tas
 		return nil, microerror.Mask(err)
 	}
 
+	args := []string{
+		"registry",
+		"push",
+	}
+	if len(projectInfo.Channels) > 0 {
+		args = append(args, fmt.Sprintf("--channel=%s", projectInfo.Channels[0]))
```
rossf7: I'm not sure we actually use multiple channels. But will this work with the csv input format?

cmd/build.go
kopiczko: I'm thinking if we can skip flags all together and check `CIRCLE_BRANCH` env var. If it's set and it's set to a different value than `master` then we skip pushing to stable channel. No need for flags at all.

fiunchinho: @rossf7 To be backwards compatible, the channel shouldn't be passed. We weren't using `stable`, we werent't passing anything.

rossf7: > the channel shouldn't be passed.

@fiunchinho Got it. 👍 

@kopiczko I like your idea to check `CIRCLE_BRANCH`. The Circle CI config should prevent this. But sometimes this doesn't work so this would be a fallback.

https://github.com/giantswarm/apiextensions/blob/5bd95e0af5b6b99187b0009d124bf8a232f920f8/.circleci/config.yml#L17-L19

kopiczko: For my understanding: where/when would you like to specify the flag?

fiunchinho: in the apiextensions repository, we don't want the unmerged PR's to overwrite the stable channel

fiunchinho: > I'm thinking if we can skip flags all together and check `CIRCLE_BRANCH` env var. If it's set and it's set to a different value than `master` then we skip pushing to stable channel. No need for flags at all.

something like
```
branch := os.Getenv("CIRCLE_BRANCH")
	if branch != "master" {
		return w, nil
	}

	helmTasks, err := processHelmDir(fs, projectInfo, NewHelmPushTask)
	if err != nil {
		return nil, microerror.Mask(err)
	}

	w = append(w, helmTasks...)

	return w, nil
```

kopiczko: This is the build config of `apiextensions` https://github.com/giantswarm/apiextensions/blob/e11354aee7703fa74c504978a3d4b525c7814730/.circleci/config.yml#L1-L19

```yaml
version: 2
jobs:
  build:
    machine: true
    steps:
    - checkout

    - run: |
        wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
        chmod +x ./architect
        ./architect version
    - run: ./architect build
    - store_test_results:
        path: /tmp/results
    - deploy:
        command: |
          if [ "${CIRCLE_BRANCH}" == "master" ]; then
            ./architect deploy
          fi
```

What is pushing to the `stable` channel on non-master builds? `architect build`?

fiunchinho: yes

fiunchinho: https://github.com/giantswarm/architect/blob/1017c7b25adeeb2d353e3f576943ae17b080b634/workflow/workflow.go#L188

kopiczko: Then your prototype from here https://github.com/giantswarm/architect/pull/359#discussion_r377718731 should be fine with one caveat. When the env var doesn't exist at all we need to preserve the old behaviour. There is non-zero chance someone uses `architect build` for local testing.

fiunchinho:
Please take a look.

workflow/workflow.go
```diff
@@ -185,12 +186,15 @@ func NewBuild(projectInfo ProjectInfo, fs afero.Fs) (Workflow, error) {
 		}
 	}
 
-	helmTasks, err := processHelmDir(fs, projectInfo, NewHelmPushTask)
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
+	branch := os.Getenv("CIRCLE_BRANCH")
```
kopiczko: I think using https://golang.org/pkg/os/#LookupEnv would be cleaner here.

workflow/workflow_test.go
```diff
@@ -243,6 +244,9 @@ func TestGetBuildWorkflow(t *testing.T) {
 			setUp: func(fs afero.Fs, testDir string) error {
 				projectInfo.WorkingDirectory = testDir
 
+				if err := os.Setenv("CIRCLE_BRANCH", ""); err != nil {
```
kopiczko: Is that really needed? By default `os.Getenv` should return `""` right?

fiunchinho: No because these tests are being ran during CI, so the env var is there

kopiczko: I see. Let's add a comment here. And maybe a `defer` to restore the old value.

fiunchinho: Comment added. Not sure how to go about the defer function, since this is done in the setup, and we don't have a teardown phase.

kopiczko: But it will restore the environment at the end of the test. This is OS-wide change if I'm not mistaken.

fiunchinho: Take a look now

workflow/workflow_test.go
```diff
@@ -379,6 +380,17 @@ func TestGetBuildWorkflow(t *testing.T) {
...
+	// Build task won't push the chart to the registry when on a branch other than master, so we need to
+	// unset the CIRCLE_BRANCH which is set while running this own repository CI.
+	currentBranch := os.Getenv("CIRCLE_BRANCH")
+	if err := os.Unsetenv("CIRCLE_BRANCH"); err != nil {
+		t.Fatalf("unexpected error during setup: %#v", err)
+	}
+
+	defer func() {
+		os.Setenv("CIRCLE_BRANCH", currentBranch)
+	}()
```
kopiczko: `defer os.Setenv("CIRCLE_BRANCH", currentBranch)`

PR #360, opened 2020-02-12
Add Helm chart testing tool

yasn77:
This commit adds Helm Chart testing tool
(https://github.com/helm/chart-testing). Rather than copy what is in the
Dockerfile of that repo, simply use the image and copy aritfacts from it. This
ensures we have exactly the same versions as used in upstream Image.

Will go towards https://github.com/giantswarm/architect-orb/pull/65

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,21 @@
+FROM quay.io/giantswarm/helm-chart-testing:v2.4.0 AS ct
+
+RUN pip freeze > /py-requirements.txt
+
 # Stage 0
-FROM quay.io/giantswarm/golang:1.13.1-alpine3.10
+FROM quay.io/giantswarm/golang:1.13.1-alpine3.10 AS golang
 
 # Stage 1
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,11 +1,21 @@
+FROM quay.io/giantswarm/helm-chart-testing:v2.4.0 AS ct
+
+RUN pip freeze > /py-requirements.txt
+
 # Stage 0
```
kopiczko:
```suggestion
```

Dockerfile
```diff
@@ -1,11 +1,21 @@
+FROM quay.io/giantswarm/helm-chart-testing:v2.4.0 AS ct
+
+RUN pip freeze > /py-requirements.txt
```
kopiczko: Can we give it some more meaningful name like `helm-chart-testing-py-requirements.txt`? I think line `RUN pip install -r /helm-chart-testing-py-requirements.txt` is more self-explanatory.

PR #361, opened 2020-02-15
Revert #359

fiunchinho:
This was a feature used to push the helm chart to the registry while working on operators branches. That made possible to deploy the operator chart to run e2e tests.

We need to find a way to make the apiextensions repository to push the helm charts using sha instead of always overwritting `0.1.0`.


fiunchinho:
Merging since this is blocking us from testing a fix on azure-operator.

PR #362, opened 2020-02-18
cmd, workflow: stop injecting version into build info

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/8637

We want to invert how the project version is maintained and instead of
generating it from git data and injecting during the build, we want to
maintain it in source files. This will allow managing releases by
merging PRs and that they'll undergo a review process.

### DIFF
--- a/workflow/golang_test.go
+++ b/workflow/golang_test.go
@@ -2,6 +2,7 @@ package workflow
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/giantswarm/architect/tasks"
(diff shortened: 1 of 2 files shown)


sslavic:
Every commit (merged squashed PR to master) is becoming a release? If so, would git commit sha be enough?

Also, how does one differentiate between different WIP build outputs (e.g. from different builds with same or additional commits) in same PR? E.g. I'd like to include and test each WIP change on a PR to an app like nginx IC App in legacy platform WIP releases, before releasing (merging to master) the app.

kopiczko:
> Every commit (merged squashed PR to master) is becoming a release? If so, would git commit sha be enough?
> 
> Also, how does one differentiate between different WIP build outputs (e.g. from different builds with same or additional commits) in same PR? E.g. I'd like to include and test each WIP change on a PR to an app like nginx IC App in legacy platform WIP releases, before releasing (merging to master) the app.

Chart versioning doesn't change. The versioning of app inside (e.g. aws-operator) changes.

workflow/golang_test.go
```diff
@@ -55,8 +55,7 @@ func TestNewGoBuildTask(t *testing.T) {
 					"-extldflags '-static' " +
 					"-X 'main.gitCommit=e8363ac222255e991c126abe6673cd0f33934ac8' " +
 					"-X 'github.com/giantswarm/architect/pkg/project.buildTimestamp=2019-06-04T12:40:05Z' " +
-					"-X 'github.com/giantswarm/architect/pkg/project.gitSHA=e8363ac222255e991c126abe6673cd0f33934ac8' " +
-					"-X 'github.com/giantswarm/architect/pkg/project.version=test'",
+					"-X 'github.com/giantswarm/architect/pkg/project.gitSHA=e8363ac222255e991c126abe6673cd0f33934ac8'",
```
kopiczko: I feel like it should be passed to `strings.Join` to avoid ugly diffs, but that's a nit.

mig4: I might go ahead and do it cause it's bitten me already (I didn't notice the space at the end and tests failed).

mig4: Done

mig4:
Going with this soon.

PR #366, opened 2020-03-03
helmtemplate: Add branch name to template context

mig4:
Towards https://github.com/giantswarm/giantswarm/issues/8632

Add Git branch name to context data used when templating Helm charts.
It's already available as a flag but previously wasn't exposed to the
template. We now want to include it to be able to expose it via a label
on the resources created by the template.

### DIFF
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -13,6 +13,7 @@ import (
 func runTemplateError(cmd *cobra.Command, args []string) error {
 	var (
 		chartDir = cmd.Flag("dir").Value.String()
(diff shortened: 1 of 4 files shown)


mig4:
> Looks sane to me - is this also in the architect-orb already?

AFAICT `architect-orb` will just use the `architect` binary for templating Helm charts so we should be good here.

mig4:
Thanks! Yip, tested locally on an `aws-operator` branch where I've added this variable to `values.yaml` and it works OK. Also, unit tests expanded to cover it here too.

PR #369, opened 2020-03-10
Retrieves version from project.go file

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/8632

Read the version stored in the project's source code at `pkg/project/project.go`.

### DIFF
--- a/cmd/helm/template/runner_test.go
+++ b/cmd/helm/template/runner_test.go
@@ -1,0 +1,82 @@
+package template
+
+import (
+	"context"
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/giantswarm/gitrepo/pkg/gitrepo"
+	repo "github.com/giantswarm/gitrepo/pkg/gitrepo"
+)
+
+// TestGetProjectVersions tests getProjectversion method which retrieves
+// the value of the version variable in pkg/project/project.go file.
+func TestGetProjectVersion(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name            string
+		repoURL         string
+		tag             string
+		expectedVersion string
+		errorMatcher    func(err error) bool
+	}{
+		{
+			name:            "case 0: aws-operator@v8.1.1",
+			repoURL:         "git@github.com:giantswarm/aws-operator.git",
+			tag:             "v8.1.1",
+			expectedVersion: "n/a",
+		},
+		{
+			name:            "case 1: azure-operator@v3.0.0",
+			repoURL:         "git@github.com:giantswarm/azure-operator.git",
+			tag:             "v3.0.0",
+			expectedVersion: "3.0.0",
+		},
+	}
+
+	for i, tc := range testCases {
+		t.Run(strconv.Itoa(i), func(t *testing.T) {
+			t.Log(tc.name)
+
+			dir := "/tmp/architect-test-getprojectversion"
+			defer os.RemoveAll(dir)
+
+			c := repo.Config{
+				Dir: dir,
+				URL: tc.repoURL,
+			}
+			repo, err := repo.New(c)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			ctx := context.Background()
+
+			err = repo.EnsureUpToDate(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			version, err := getProjectVersion(repo, gitrepo.CheckoutOptions{Tag: tc.tag})
+
+			switch {
+			case err == nil && tc.errorMatcher == nil:
+				// correct; carry on
+			case err != nil && tc.errorMatcher == nil:
+				t.Fatalf("error == %#v, want nil", err)
+			case err == nil && tc.errorMatcher != nil:
+				t.Fatalf("error == nil, want non-nil")
+			case !tc.errorMatcher(err):
+				t.Fatalf("error == %#v, want matching", err)
+			}
+
+			if err == nil {
+				if version != tc.expectedVersion {
+					t.Errorf("got %#q, expected %#q\n", version, tc.expectedVersion)
+				}
+			}
+		})
+	}
+}
--- a/cmd/helm/template/runner_test.go
+++ b/cmd/helm/template/runner_test.go
@@ -6,7 +6,6 @@ import (
 	"strconv"
 	"testing"
 
-	"github.com/giantswarm/gitrepo/pkg/gitrepo"
 	repo "github.com/giantswarm/gitrepo/pkg/gitrepo"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/hook/pre_run_e.go
```diff
@@ -49,15 +53,71 @@ func PreRunE(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return microerror.Mask(err)
 	}
 
 	cmd.PersistentFlags().String("branch", defaultBranch, "git branch being built")
 	cmd.PersistentFlags().String("sha", defaultSha, "git SHA1 being built")
 	cmd.PersistentFlags().String("tag", defaultTag, "git tag being built")
-	cmd.PersistentFlags().String("version", defaultVersion, "project version being built")
+	cmd.PersistentFlags().String("version", gitVersion, "version found in git")
+	cmd.PersistentFlags().String("source-version", srcVersion, "version found in source code")
```
TheoBrigitte: Could not think of a better name, suggestions welcome.

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/giantswarm/apiextensions v0.0.0-20191005094604-efa5c38da056 // indirect
 	github.com/giantswarm/app v0.0.0-20200108121642-98786417f1e1
-	github.com/giantswarm/gitrepo v0.0.0-20191202170843-422800b3c4bb // master
+	github.com/giantswarm/gitrepo v0.0.0-20200310120849-544dc51a1302 // master
```
TheoBrigitte: Reminder: after merging gitrepo PR run `go mod edit -require=github.com/giantswarm/gitrepo@master`

cmd/hook/pre_run_e.go
```diff
@@ -49,15 +53,71 @@ func PreRunE(cmd *cobra.Command, args []string) error {
...
+// parseString returns the value of the variable varName found in r AbstractSyntaxTree.
+func parseString(r ast.Node, varName string) (value string, err error) {
+	ast.Inspect(r, func(n ast.Node) bool {
+		d, ok := n.(*ast.ValueSpec)
+		if ok {
+			for _, id := range d.Names {
+				if id.Name == varName {
+					v := id.Obj.Decl.(*ast.ValueSpec).Values[0].(*ast.BasicLit)
+					if v.Kind == token.STRING {
+						value, err = strconv.Unquote(v.Value)
```
xh3b4sd: I think we should stop here somehow. Maybe by returning `false`. Otherwise eventual errors could potentially be overwritten again.

cmd/hook/pre_run_e.go
xh3b4sd: So is it "git", "app" and "src" version we have?

cmd/hook/pre_run_e.go
kopiczko: Or assert that we have only one id with name set to varName.

go.mod
kopiczko: Actually it would be good to release gitrepo.

cmd/hook/pre_run_e.go
```diff
@@ -49,15 +53,71 @@ func PreRunE(cmd *cobra.Command, args []string) error {
...
 }
+
+// getProjectVersion retrieves version stored in project's Go source code.
+// It looks up the value of variable `version` in `pkg/project/project.go` file
+// on master branch.
+func getProjectVersion(repo *gitrepo.Repo) (string, error) {
+	filePath := "pkg/project/project.go"
+	varName := "version"
+
+	content, err := repo.GetFileContent(filePath)
```
kopiczko: There are projects like managed apps which don't have this file. We need to handle that as well. I believe having this code in the hook will make architect fail for this kind of projects.

cmd/hook/pre_run_e.go
kopiczko: In helm lingo it would be app version right?

rossf7: Yes in Helm it's called `appVersion`. In releases we call it `componentVersion`.

https://github.com/giantswarm/releases/blob/277f0a2999d3284ca685a83935586952e991bb95/aws.yaml#L9

I'd prefer we use `app-version`. Maybe we could inject it into the `Chart.yaml` somehow? If so it will also appear in helm CLI output.

kopiczko: The whole purpose is to inject it into `appVersion` of `Chart.yaml`. So yeah that makes sense in my head now. I'm also voting for `app-version`.

cmd/hook/pre_run_e.go
kopiczko: In that case we should require `appVersion` to be present in `Chart.yaml`. That makes me think we shouldn't add it to the hook but to the command itself.

mig4: Oh as in if that returns an error we read in `Chart.yaml` and scream if `appVersion` is not there or doesn't have a static value?

What value should it have in e.g. chart-only apps like the `gatekeeper-control-plane-rules`? Wouldn't it just require people to manually set it to the same thing as `version`?

kopiczko: > Oh as in if that returns an error we read in `Chart.yaml` and scream if `appVersion` is not there or doesn't have a static value?

Yes that's my thinking.

> What value should it have in e.g. chart-only apps like the `gatekeeper-control-plane-rules`? Wouldn't it just require people to manually set it to the same thing as `version`?

Yeah I think `appVersion: [[ .Version ]]` makes sense there.

kopiczko: It should also scream if you have `pkg/project/project.go` defined and `appVersion` set in your `Chart.yaml`.

cmd/hook/pre_run_e.go
TheoBrigitte: ok let's go with `app-version` it make total sense.

cmd/hook/pre_run_e.go
TheoBrigitte: added `return false` to stop execution

cmd/hook/pre_run_e.go
TheoBrigitte: I moved this logic in the `cmd/helm/template` command. Also I would like to handle what we do when with versions in a separate PR.

cmd/helm/template/runner_test.go
```diff
@@ -0,0 +1,82 @@
...
+		repoURL         string
+		tag             string
+		expectedVersion string
+		errorMatcher    func(err error) bool
+	}{
+		{
+			name:            "case 0: aws-operator@v8.1.1",
+			repoURL:         "git@github.com:giantswarm/aws-operator.git",
+			tag:             "v8.1.1",
+			expectedVersion: "n/a",
```
kopiczko: Where does this `"n/a"` come from?

TheoBrigitte: right there > https://github.com/giantswarm/aws-operator/blob/1ec7d04f7f3f3ede11adc23441681ec6fa4d2075/pkg/project/project.go#L9

PR #371, opened 2020-03-17
Provide .AppVersion as Chart.yaml value

wtty-fool:
Towards: giantswarm/giantswarm#8632

Adds version from `project.go` as `AppVersion` value for the chart.

### DIFF
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -55,12 +55,12 @@ func runTemplateError(cmd *cobra.Command, args []string) (err error) {
 	var s *helmtemplate.TemplateHelmChartTask
 	{
 		c := helmtemplate.Config{
-			Fs:           fs,
-			ChartDir:     chartDir,
-			Branch:       branch,
-			Sha:          sha,
-			ChartVersion: version,
-			AppVersion:   appVersion,
+			Fs:         fs,
+			ChartDir:   chartDir,
+			Branch:     branch,
+			Sha:        sha,
+			Version:    version,
+			AppVersion: appVersion,
 		}
 
 		s, err = helmtemplate.NewTemplateHelmChartTask(c)
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -18,12 +18,25 @@ import (
 
 func runTemplateError(cmd *cobra.Command, args []string) (err error) {
 	var (
-		chartDir = cmd.Flag("dir").Value.String()
-		branch   = cmd.Flag("branch").Value.String()
-		sha      = cmd.Flag("sha").Value.String()
-		tag      = cmd.Flag("tag").Value.String()
-		version  = cmd.Flag("version").Value.String()
+		chartDir    = cmd.Flag("dir").Value.String()
+		branch      = cmd.Flag("branch").Value.String()
+		sha         = cmd.Flag("sha").Value.String()
+		tag         = cmd.Flag("tag").Value.String()
+		version     = cmd.Flag("version").Value.String()
+		validate    bool
+		taggedBuild bool
 	)
+	{
+		var err error
+		validate, err = strconv.ParseBool(cmd.Flag("validate").Value.String())
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		taggedBuild, err = strconv.ParseBool(cmd.Flag("tagged-build").Value.String())
+		if err != nil {
+			return microerror.Mask(err)
+		}
+	}
 
 	fs := afero.NewOsFs()
 	ctx := context.Background()
@@ -69,7 +82,7 @@ func runTemplateError(cmd *cobra.Command, args []string) (err error) {
 		}
 	}
 
-	if err := s.Run(); err != nil {
+	if err := s.Run(validate, taggedBuild); err != nil {
 		return microerror.Mask(err)
 	}
 
--- a/cmd/helm/template/runner.go
+++ b/cmd/helm/template/runner.go
@@ -55,11 +55,12 @@ func runTemplateError(cmd *cobra.Command, args []string) (err error) {
 	var s *helmtemplate.TemplateHelmChartTask
 	{
 		c := helmtemplate.Config{
-			Fs:       fs,
-			ChartDir: chartDir,
-			Branch:   branch,
-			Sha:      sha,
-			Version:  version,
+			Fs:           fs,
+			ChartDir:     chartDir,
+			Branch:       branch,
+			Sha:          sha,
+			ChartVersion: version,
+			AppVersion:   appVersion,
 		}
 
 		s, err = helmtemplate.NewTemplateHelmChartTask(c)
--- a/helmtemplate/helm.go
+++ b/helmtemplate/helm.go
@@ -4,11 +4,11 @@ import (
 	"bytes"
 	"fmt"
 	"path"
-	"regexp"
 	"text/template"
 
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
+	"sigs.k8s.io/yaml"
 )
 
 const (
@@ -22,7 +22,6 @@ const (
 	HelmTemplateDirectoryName = "templates"
 	// keyValuePattern helps build regexp patterns which fit YAML's key-value
 	// formatting for string values
-	keyValuePattern = `%s:\s*["']?(%s)["']?`
 )
 
 // TemplateHelmChartTask is used to run a template-helm-chart command
@@ -47,6 +46,13 @@ type Config struct {
 	AppVersion string
 }
 
+// renderedChart is used for chart validation after it has been filled with
+// values
+type renderedChart struct {
+	Version    string `json:"version,omitempty"`
+	AppVersion string `json:"appVersion,omitempty"`
+}
+
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 func (t TemplateHelmChartTask) Run(validate, taggedBuild bool) error {
 	for _, file := range []string{HelmChartYamlName, HelmValuesYamlName} {
@@ -56,7 +62,8 @@ func (t TemplateHelmChartTask) Run(validate, taggedBuild bool) error {
 			return microerror.Mask(err)
 		}
 
-		if validate && taggedBuild && t.chartVersion != t.appVersion {
+		// We expect versions to match for a tagged build if project.go has been found.
+		if validate && taggedBuild && t.appVersion != "" && t.chartVersion != t.appVersion {
 			return microerror.Newf(
 				"version in the tag does not match version defined in project.go: %q != %q",
 				t.chartVersion, t.appVersion,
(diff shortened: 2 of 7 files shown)


wtty-fool:
WIP: still need to add validation

cmd/helm/template/runner.go
```diff
@@ -55,11 +55,12 @@ func runTemplateError(cmd *cobra.Command, args []string) (err error) {
...
-			Fs:       fs,
-			ChartDir: chartDir,
-			Branch:   branch,
-			Sha:      sha,
-			Version:  version,
+			Fs:           fs,
+			ChartDir:     chartDir,
+			Branch:       branch,
+			Sha:          sha,
+			ChartVersion: version,
```
TheoBrigitte: I would like to keep the variable name `Version` on the left hand side here. As this align with nicely with chart, where the chart version is just called version.

helmtemplate/types.go
```diff
@@ -18,4 +18,6 @@ type BuildInfo struct {
 	SHA string
 	// Version is the version of the commit being built.
 	Version string
+	// AppVersion is the version read from project.go
```
kopiczko:
```suggestion
	// AppVersion is the version read from pkg/project/project.go if it exists or set to the same value as Version otherwise.
```

helmtemplate/helm_test.go
```diff
@@ -12,27 +12,29 @@ import (
...
+			appVersion:   "1.0.0",
 			setUp: func(fs afero.Fs, chartDir string) error {
 				files := []struct {
 					path string
 					data string
 				}{
 					{
 						path: filepath.Join(chartDir, HelmChartYamlName),
-						data: "version: 1.0.0-[[ .SHA ]]",
+						data: "version: [[ .AppVersion ]]-[[ .SHA ]]",
```
mig4: This doesn't seem quite right, but probably because it wasn't right before, we should be maybe testing `[[ .Version ]]` template as well but it was hardcoded as `1.0.0`, I'd probably do:

```
data: "version: [[ .Version ]]\nappVersion: [[ .AppVersion ]]"
```

so it's clearer, and possibly move the `[[ .SHA ]]` part to the test case below.

wtty-fool: Thanks for the suggestion. I did just that.

cmd/helm/template/runner.go
wtty-fool: Yup, done.

helmtemplate/helm.go
```diff
@@ -66,15 +80,35 @@ func (t TemplateHelmChartTask) Run() error {
...
 		if err := afero.WriteFile(t.fs, path, buf.Bytes(), permission); err != nil {
 			return microerror.Mask(err)
 		}
 	}
 	return nil
 }
 
+// ValidateChart makes sure version fields and values made it into the
+// chart when executing the template.
+func ValidateChart(version, appVersion string, chart bytes.Buffer) error {
```
wtty-fool: Note: This function needs some serious testing (or a better way to perform this check).

helmtemplate/helm.go
```diff
@@ -19,41 +20,54 @@ const (
...
 		if err != nil {
 			return microerror.Mask(err)
 		}
 
+		if validate && taggedBuild && t.chartVersion != t.appVersion {
+			return microerror.Newf(
+				"version in the tag does not match version defined in project.go: %q != %q",
+				t.chartVersion, t.appVersion,
+			)
+		}
```
mig4: This could probably be done before the loop.

helmtemplate/helm.go
mig4: For me it would be preferable to load the YAML instead of using regex, WDYT?

helmtemplate/helm.go
```diff
@@ -99,12 +133,17 @@ func NewTemplateHelmChartTask(config Config) (*TemplateHelmChartTask, error) {
 		return nil, microerror.Maskf(invalidConfigError, "%T.Version must not be empty", config)
 	}
 
+	if config.AppVersion == "" {
+		return nil, microerror.Maskf(invalidConfigError, "%T.AppVersion must not be empty", config)
+	}
+
```
TheoBrigitte: This test is not needed `AppVersion` can be empty.

helmtemplate/helm.go
wtty-fool: I do agree, however logic changed following today's discussion.

wtty-fool:
I think it's ready for your reviews, chaps <3

helmtemplate/helm.go
```diff
@@ -66,15 +86,47 @@ func (t TemplateHelmChartTask) Run() error {
...
+	chart := renderedChart{}
+	err := yaml.Unmarshal(chartBuf.Bytes(), &chart)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	if chart.Version != version {
+		return microerror.Maskf(
+			versionValidationError,
+			"validation error: chart does not contain \"version\": %q", version,
```
TheoBrigitte: No need to repeat `validation error` as its already contained in `versionValidationError` error.

```suggestion
			"wrong value for \"version\" in chart: got %#q, expect %#q", chart.Version, version,
```

helmtemplate/helm.go
```diff
@@ -66,15 +86,47 @@ func (t TemplateHelmChartTask) Run() error {
...
+		)
+	}
+
+	// We want to validate appVersion only when the project.go has been found,
+	// i.e. appVersion is non-empty.
+	if appVersion != "" && chart.AppVersion != appVersion {
+		return microerror.Maskf(
+			versionValidationError,
+			"validation error: chart does not contain \"appVersion\": %q", appVersion,
```
TheoBrigitte:
```suggestion
			"wrong value for \"appVersion\" in chart: got %#q, expect %#q", chart.AppVersion, appVersion,
```

helmtemplate/helm.go
```diff
@@ -66,15 +86,47 @@ func (t TemplateHelmChartTask) Run() error {
 			return microerror.Mask(err)
 		}
 
+		if validate && ValidateChart(buildInfo.Version, buildInfo.AppVersion, buf) != nil {
```
TheoBrigitte: This will fail on the second iteration when validating the `values.yaml` file aka  `HelmValuesYamlName`.

cmd/helm/template/runner.go
```diff
@@ -18,12 +18,25 @@ import (
...
+		branch      = cmd.Flag("branch").Value.String()
+		sha         = cmd.Flag("sha").Value.String()
+		tag         = cmd.Flag("tag").Value.String()
+		version     = cmd.Flag("version").Value.String()
+		validate    bool
+		taggedBuild bool
 	)
+	{
+		var err error
+		validate, err = strconv.ParseBool(cmd.Flag("validate").Value.String())
```
kopiczko: Isn't it possible to do something like `cmd.Flag("version").Value.Bool()`?

helmtemplate/helm.go
```diff
@@ -25,24 +26,41 @@ const (
...
+	Version    string
+	AppVersion string
+}
+
+// renderedChart is used for chart validation after it has been filled with
+// values
+type renderedChart struct {
+	Version    string `json:"version,omitempty"`
+	AppVersion string `json:"appVersion,omitempty"`
 }
```
kopiczko: Let's please move this to `types.go`.

helmtemplate/helm.go
```diff
@@ -25,24 +26,41 @@ const (
...
 
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
-func (t TemplateHelmChartTask) Run() error {
+func (t TemplateHelmChartTask) Run(validate, taggedBuild bool) error {
+	// We expect versions to match for a tagged build if project.go has been found.
+	if validate && taggedBuild && t.appVersion != "" && t.chartVersion != t.appVersion {
```
kopiczko: I'd drop a comment that `t.appVersion ` is empty when `pkg/project/project.go` file is not found.

helmtemplate/helm.go
```diff
@@ -25,24 +26,41 @@ const (
...
+		return microerror.Maskf(
+			versionMismatchError,
```
kopiczko: As fmt says. If matcher is not used then it should be `executionFailedError`.

```suggestion
			executionFailedError,
```

helmtemplate/helm.go
```diff
@@ -66,15 +85,49 @@ func (t TemplateHelmChartTask) Run() error {
...
 		if err := afero.WriteFile(t.fs, path, buf.Bytes(), permission); err != nil {
 			return microerror.Mask(err)
 		}
 	}
 	return nil
 }
 
+// ValidateChart makes sure version fields and values made it into the
+// chart when executing the template.
+func ValidateChart(version, appVersion string, chartBuf bytes.Buffer) error {
```
kopiczko:
```suggestion
func validateChart(version, appVersion string, chartBuf bytes.Buffer) error {
```

helmtemplate/types.go
```diff
@@ -10,6 +10,13 @@ var (
 	permission os.FileMode = 0644
 )
 
+const (
+	BranchTag     = "[[ .Branch ]]"
+	SHATag        = "[[ .SHA ]]"
+	VersionTag    = "[[ .Version ]]"
+	AppVersionTag = "[[ .AppVersion ]]"
+)
```
kopiczko: Why do we have here stuff used in tests? Also maybe golden style tests make sense here? https://github.com/giantswarm/devctl/blob/e727bcf6eba47e67f387bd6ff21fb4c2e1763e01/pkg/gen/resource/golden_test.go#L28

cmd/helm/template/runner.go
wtty-fool: We would have to use global vars.
```go
var validate, taggedBuild bool
...
Cmd.Flags().BoolP(&validate, ...)
Cmd.Flags().BoolP(&taggedBuild, ...)
```

kopiczko: Okay let's leave like that for now. Eventually we should switch `devctl`-like structure.

helmtemplate/types.go
wtty-fool: Golden style tests make sense, but I'll let @TheoBrigitte decide, since he volunteered for that :)
Just wanted to extract those strings to avoid misspelling them in tests.

helmtemplate/types.go
```diff
@@ -18,4 +18,14 @@ type BuildInfo struct {
 	SHA string
 	// Version is the version of the commit being built.
 	Version string
+	// AppVersion is the version read from pkg/project/project.go if it
+	// exsitsts or set to the same value as Version otherwise.
```
TheoBrigitte:
```suggestion
	// exists or set to the same value as Version otherwise.
```

helmtemplate/helm.go
```diff
@@ -66,15 +80,49 @@ func (t TemplateHelmChartTask) Run() error {
 			return microerror.Mask(err)
 		}
 
+		if file == HelmChartYamlName && validate && validateChart(buildInfo.Version, buildInfo.AppVersion, buf) != nil {
+			return microerror.Mask(err)
```
kopiczko: What is the error you're masking here? I think you wanted to mask what is returned by `validateChart` right?

PR #374, opened 2020-03-19
template: read project.go from disk

kopiczko:


cmd/helm/template/runner.go
```diff
@@ -94,11 +87,12 @@ func runTemplateError(cmd *cobra.Command, args []string) (err error) {
 // getProjectVersion retrieves version stored in project's Go source code.
 // It looks up the value of variable `version` in `pkg/project/project.go` file
 // on version defined in ref.
-func getProjectVersion(repo *gitrepo.Repo, ref string) (string, error) {
+func getProjectVersion(ctx context.Context, repoDir string) (string, error) {
 	filePath := "pkg/project/project.go"
 	varName := "version"
 
-	content, err := repo.GetFileContent(filePath, ref)
```
kopiczko: @mig4 found out this was doing "something bad" when `architect helm template` was executed from a branch. This is because this was silently checking out to "origin/master". I think using git get get this file content is overkill. We can simply read it from FS.

cmd/helm/template/runner_test.go
```diff
@@ -22,43 +23,23 @@ func TestGetProjectVersion(t *testing.T) {
...
-			tag:             "v8.1.1",
-			expectedVersion: "n/a",
-		},
-		{
-			name:            "case 1: azure-operator@v3.0.0",
-			repoURL:         "git@github.com:giantswarm/azure-operator.git",
-			tag:             "v3.0.0",
-			expectedVersion: "3.0.0",
+			name:            "case 0",
+			expectedVersion: project.Version(),
```
kopiczko: Those tests were taking long and because of that were unreliable.

cmd/build.go
```diff
@@ -41,7 +41,7 @@ func init() {
 	var defaultDockerUsername string
 	var defaultDockerPassword string
 
-	if os.Getenv("CIRCLECI") == "true" {
+	if os.Getenv("CIRCLECI") == "true" { // nolint:goconst
```
TheoBrigitte: What is this for ?

template/helm.go
```diff
@@ -40,14 +40,14 @@ type TemplateHelmChartTask struct {
 
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 func (t TemplateHelmChartTask) Run() error {
-	err := afero.Walk(t.fs, t.chartDir, func(path string, info os.FileInfo, err error) error {
+	err := afero.Walk(t.fs, t.chartDir, func(path string, info os.FileInfo, _ error) error {
```
TheoBrigitte: This error should be handled.

workflow/workflow_test.go
```diff
@@ -387,7 +387,7 @@ func TestGetBuildWorkflow(t *testing.T) {
 		if err != nil {
 			t.Fatalf("test %d received unexpected error creating a test directory: %#v", i, err)
 		}
-		defer fs.RemoveAll(testDir)
+		defer fs.RemoveAll(testDir) //nolint:errcheck
```
TheoBrigitte: What does this do?

kopiczko: Linter complained about non-checked error returned by RemoveAll.

template/helm.go
kopiczko: That's true. But it wasn't before

cmd/build.go
kopiczko: Linter complained that `"true"` is there for 5 times in the code and that should be a const. and I disagree :D

TheoBrigitte: Is this linter from ci ?

kopiczko: https://circleci.com/gh/giantswarm/architect/1238?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #375, opened 2020-03-19
circleci: update architect orb to v0.8.3

kopiczko:


cmd/build.go
```diff
@@ -41,7 +41,7 @@ func init() {
 	var defaultDockerUsername string
 	var defaultDockerPassword string
 
-	if os.Getenv("CIRCLECI") == "true" {
+	if _, ok := os.LookupEnv("CIRCLECI"); ok {
```
kopiczko: We need to check if the value is `"true"` because that's what's defined by the circleci https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

kopiczko: Linter says:

```
cmd/build.go:44:30: string `true` has 5 occurrences, make it a constant (goconst)
```

So we need to do something like `if os.Getenv("CIRCLECI") == trueString {`...

kopiczko: Or rather:

```suggestion
	if os.Getenv("CIRCLECI") == "true" { //nolint:goconst
```

wtty-fool: Sure. I thought it's enough to check whether `CIRCLECI` is there at all, but I guess we'll avoid accidental corner case of `$CIRCLECI=false`.

kopiczko:
Can't approve PR I opened so I'll merge instead :D

PR #379, opened 2020-03-19
Refactor test for TemplateHelmChartTask

TheoBrigitte:
Towards: giantswarm/giantswarm#8632

* Refactor test to use golden files/folder.
* Add more test cases for validation

### DIFF
--- a/helmtemplate/helm_test.go
+++ b/helmtemplate/helm_test.go
@@ -1,11 +1,14 @@
 package helmtemplate
 
 import (
+	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/giantswarm/microerror"
+	"github.com/google/go-cmp/cmp"
 	"github.com/spf13/afero"
 )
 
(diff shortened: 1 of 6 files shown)


helmtemplate/helm_test.go
```diff
@@ -1,127 +1,200 @@
 package helmtemplate
 
 import (
+	"bytes"
 	"fmt"
+	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/giantswarm/microerror"
```
wtty-fool: NIT: Maybe you could separate 3rd party libraries from our own in imports.

helmtemplate/helm_test.go
```diff
@@ -1,127 +1,200 @@
...
-					{
-						path: filepath.Join(chartDir, HelmChartYamlName),
-						data: fmt.Sprintf("version: %s\nappVersion: %s", VersionTag, AppVersionTag),
-					},
-					{
-						path: filepath.Join(chartDir, HelmValuesYamlName),
-						data: fmt.Sprintf("branch: %s\ncommit: %s", BranchTag, SHATag),
-					},
-				}
+			name: "case 0: chart templating",
```
wtty-fool: SUPER-NIT:
I don't know how important the case names are, but the descriptions are a bit lacking.
Especially for the cases we expect to fail, it would be nice to know why do we expect them to fail, since the error will be caught by the test itself.

If you feel this is to much info to cram into case name, maybe a comment would do?

helmtemplate/helm_test.go
TheoBrigitte: Those imports looks correct according to our guide https://github.com/giantswarm/fmt/blob/master/go/imports.md

helmtemplate/helm_test.go
TheoBrigitte: I agree, it would be good to give more information on what is this test doing, also I do not want the test name to become too long so a comment sounds good. Do you have a suggestion maybe?

TheoBrigitte: Changed test name to
```
+                       name: "case 0: template variables",
+                       name: "case 1: validate version == appversion",
+                       name: "case 2: version validation failure",
+                       name: "case 3: appversion validation failure",
+                       name: "case 4: version == appversion validation failure",
```

helmtemplate/helm_test.go
wtty-fool: Oh, ok then. I have to re-read the whole `go` section of fmt to avoid those kinds of comments. Carry on.

helmtemplate/helm_test.go
wtty-fool: It does, it tells me what a particular case is trying to check. Those names look better to me.

As for the comments, I've changed my mind. I think there is enough information in [giantswarm/fmt](https://github.com/giantswarm/fmt) and in comments in code. There is no point in explaining versions the 3rd time in testing code :man_shrugging:

PR #382, opened 2020-03-19
template: fix helm directory walking

kopiczko:


template/helm.go
```diff
@@ -41,6 +41,9 @@ type TemplateHelmChartTask struct {
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 func (t TemplateHelmChartTask) Run() error {
 	err := afero.Walk(t.fs, t.chartDir, func(path string, info os.FileInfo, _ error) error {
```
TheoBrigitte: The incoming error should not be ignored.

kopiczko: Yes. But I'm trying to fix immediate issue quickly.

PR #389, opened 2020-03-23
template: rename --tagged-build to --tag-build

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/8635

I realised there are "branch" builds and "tag" builds and this makes so
much more sense in my head.


kopiczko:
@wtty-fool PTALA

PR #394, opened 2020-03-29
Install conftest and policies

fiunchinho:
Installing conftest would allow us to check our kubernetes manifests / helm charts using OPA policies.

PoC

image

When changing `Deployment` api version

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN pip freeze > /helm-chart-testing-py-requirements.txt
 
 FROM quay.io/giantswarm/golang:1.13.1-alpine3.10 AS golang
 
-FROM instrumenta/conftest:v0.18.1 AS conftest
+FROM quay.io/giantswarm/conftest:v0.18.1 AS conftest
 
 # Build Image
 FROM quay.io/giantswarm/alpine:3.10
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,8 @@ RUN pip freeze > /helm-chart-testing-py-requirements.txt
 
 FROM quay.io/giantswarm/golang:1.13.1-alpine3.10 AS golang
 
+FROM instrumenta/conftest:v0.18.1 AS conftest
+
 # Build Image
 FROM quay.io/giantswarm/alpine:3.10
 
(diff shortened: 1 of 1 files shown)


yasn77:
After looking at chart testing, i felt that reusing existing Docket images (if possible) was a decent idea. This way the image itself can be used without a direct link to `archiect`, also we don't bake in too much stuff in to architect, i.e we don't have large Dockerfile with lots of build steps for various tools..  It might be worth creating separate contest Docker image for conftest and using `FROM` build steps (if possible) in architect.

MarcelMue:
What happens if no test is defined? Where are tests defined? Am I being a dummy here?

fiunchinho:
This PR only installs the `conftest` tool. It won't do anything yet.

I have [another PR](https://github.com/giantswarm/architect-orb/pull/112/files) where I used the installed `conftest` tool. In that other PR, I update our `architect-orb` to use the k8s deprecation policies (you can see the policies as _tests) with `conftest` so that it checks whether or not our Helm charts contain things that will be deprecated soon.

Dockerfile
```diff
@@ -4,6 +4,8 @@ RUN pip freeze > /helm-chart-testing-py-requirements.txt
 
 FROM quay.io/giantswarm/golang:1.13.1-alpine3.10 AS golang
 
+FROM instrumenta/conftest:v0.18.1 AS conftest
```
yasn77: I think it's worth using retagger and putting the image in our registries

TheoBrigitte:
I feel like this `conftest ` tool could live happily in its own job/command in the https://github.com/giantswarm/architect-orb/ repo and do not necessarily need to be packaged here with `architect`

fiunchinho:
> I feel like this `conftest ` tool could live happily in its own job/command in the https://github.com/giantswarm/architect-orb/ repo and do not necessarily need to be packaged here with `architect`

I thought I needed to package `conftest` here to be able to use it like [this](https://github.com/giantswarm/architect-orb/pull/112/files#diff-f063df7925141bef44a18a35d3e9814cR9). Is there a better way?

fiunchinho:
Please, take a look again. Thanks.

PR #399, opened 2020-04-09
dockerfile: don't install ct

kopiczko:
It's used as dockerized version in architect orb.

See https://github.com/giantswarm/architect-orb/pull/125.


kopiczko:
I just realised I can't merge it because it will break previous versions of the orb. We need to sort https://github.com/giantswarm/giantswarm/issues/9487 first.

kopiczko:
Closing, see https://github.com/giantswarm/architect-orb/pull/125#issuecomment-611543927.

PR #421, opened 2020-05-04
Skip AppVersion check when project file is absent

yasn77:
Managed apps do not have any go project files and the AppVersion in this
case refers to the version deployed by Helm. Not linked to any particular
project.

If project.go is not present or when checking for appVersion
in the project files returns an empty string then skip the validation checks.

Slack thread related to this commit:
https://gigantic.slack.com/archives/CRZN9GLJW/p1588605374410900

### DIFF
--- a/helmtemplate/helm.go
+++ b/helmtemplate/helm.go
@@ -26,22 +26,24 @@ const (
 type TemplateHelmChartTask struct {
 	fs afero.Fs
 
-	chartDir     string
-	branch       string
-	sha          string
-	chartVersion string
-	appVersion   string
+	chartDir            string
+	branch              string
+	sha                 string
+	chartVersion        string
+	appVersion          string
+	skipAppVersionCheck bool
 }
 
 // Config holds configuration for building a new TemplateHelmChartTask
 type Config struct {
 	Fs afero.Fs
 
-	ChartDir   string
-	Branch     string
-	Sha        string
-	Version    string
-	AppVersion string
+	ChartDir            string
+	Branch              string
+	Sha                 string
+	Version             string
+	AppVersion          string
+	SkipAppVersionCheck bool
 }
 
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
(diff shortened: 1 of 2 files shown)


helmtemplate/helm.go
```diff
@@ -26,30 +26,32 @@ const (
...
+	Version             string
+	AppVersion          string
+	SkipAppVersionCheck bool
 }
 
 // Run templates the chart's Chart.yaml and templates/deployment.yaml.
 func (t TemplateHelmChartTask) Run(validate, tagBuild bool) error {
 	// We expect versions to match for a tagged build if pkg/project/project.go
 	// file has been found. Otherwise (project.go not found) t.appVersion will
 	// be empty.
```
wtty-fool, line 53: This comment needs an update, since `appVersion` will never be empty.

helmtemplate/helm.go
```diff
@@ -26,30 +26,32 @@ const (
...
-	if validate && tagBuild && t.appVersion != "" && t.chartVersion != t.appVersion {
+	if validate && tagBuild && t.appVersion != "" && t.chartVersion != t.appVersion && !t.skipAppVersionCheck {
```
wtty-fool, line 54:
```suggestion
	if validate && tagBuild && t.chartVersion != t.appVersion && !t.skipAppVersionCheck {
```
Checking `appVersion` was supposed to catch projects with defined `project.go`, which [my change](https://github.com/giantswarm/architect/pull/421/files#diff-715449e8d707a1e6bc751c2a05943889R61) broke. So it's not necessary anymore.

helmtemplate/helm.go
yasn77: I've added a new commit that updates the comment

helmtemplate/helm.go
yasn77: At the moment, the ratio of apps being built that have `project.go` file against apps that don't, probably means having a check is valid. So @wtty-fool I think your change is still valid, just that we need to ensure apps that don't have `project.go` can still be correctly templated.

PR #422, opened 2020-05-06
Added prepare release command

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/8335

This command:
- removes the `-dev` from the `project.go` file
- adds new release to Changelog, using the version found in the `project.go` file 

With this approach, there is no arguments. The version to be released is automatically detected, based on what's in `project.go`. Same for the repository name (org/project), which uses current folder name.


cmd/release/prepare/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	"strings"
+	"time"
+
+	"github.com/giantswarm/microerror"
+	"github.com/spf13/cobra"
+)
+
+const (
+	ChangelogFile = "CHANGELOG.md"
+	VersionFile   = "pkg/project/project.go"
```
kopiczko, line 18: Why not private?

cmd/release/prepare/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	filecontents, err := ioutil.ReadFile(file)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+	versionFileContents := string(filecontents)
+
+	versionRegex := regexp.MustCompile(`version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)-dev"`)
+	currentVersion := versionRegex.FindSubmatch(filecontents)
+	if len(currentVersion) < 1 {
+		return "", microerror.Maskf(executionFailedError, "No version was found")
```
kopiczko, line 49: Error annotations should start with lower letter:

```suggestion
		return "", microerror.Maskf(executionFailedError, "no version was found")
```

cmd/release/prepare/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/spf13/cobra"
+)
+
+const (
+	ChangelogFile = "CHANGELOG.md"
+	VersionFile   = "pkg/project/project.go"
+)
+
+func runReleaseError(cmd *cobra.Command, args []string) error {
+	absoluteCurrentFolder, _ := filepath.Abs(filepath.Dir(os.Args[0]))
```
kopiczko, line 22: Instead of using arg I'd use [gitrepo.TopLevel](https://github.com/giantswarm/gitrepo/blob/32a22456f8d904d77ac94ebdfa619c1ea2fe6754/pkg/gitrepo/funcs.go#L16) for the current working directory. You can see how it's used in `hook.PreRunE` IIRC.

cmd/release/prepare/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	if !strings.Contains(changelogContents, search) {
+		return microerror.Maskf(executionFailedError, "No '[Unreleased]' work was found")
+	}
+
+	// Add new entry to the top of the changelog
+	replaceWith := fmt.Sprintf("## [Unreleased]\n\n## [%s] %s", currentVersion, date)
+	updatedFileContents := strings.Replace(changelogContents, search, replaceWith, 1)
+
+	// Update links at the bottom of the changelog
+	bottomLinks := regexp.MustCompile(`(\[Unreleased]:)(.*)(v[0-9]+\.[0-9]+\.[0-9]+)(...HEAD)\n`)
```
kopiczko, line 80: What if this is the first release?

cmd/release/prepare/runner.go
```diff
@@ -0,0 +1,89 @@
...
+		return "", microerror.Mask(err)
+	}
+	versionFileContents := string(filecontents)
+
+	versionRegex := regexp.MustCompile(`version\s*=\s*"([0-9]+\.[0-9]+\.[0-9]+)-dev"`)
+	currentVersion := versionRegex.FindSubmatch(filecontents)
+	if len(currentVersion) < 1 {
+		return "", microerror.Maskf(executionFailedError, "No version was found")
+	}
+	updatedFileContents := versionRegex.ReplaceAllString(versionFileContents, "$1")
```
kopiczko, line 51: I'd like this command to take the desired version argument because:

- validation will be easier in the workflow
- we may want to jump from `1.0.1-dev` to `1.1.0`

fiunchinho:
It needs [this merged](https://github.com/giantswarm/template-operator/pull/19) to be able to work on operators just created from template.

cmd/release/command.go
```diff
@@ -9,7 +9,7 @@ import (
 var (
 	Cmd = &cobra.Command{
 		Use:     "release",
-		Short:   "release chart as github release",
+		Short:   "release operator versions or charts",
```
kopiczko, line 12: Should be reverted?

cmd/prepare/command.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/architect/cmd/hook"
+)
+
+var (
+	Cmd = &cobra.Command{
+		Use:     "prepare-release",
```
kopiczko, line 11: It would be good to make package on par with to command so one extra package and child command like `prepare release` would be less confusing. But I understand we want to move that anyway so no blocking.

cmd/release/command.go
fiunchinho: Fixed

PR #426, opened 2020-05-07
fix cwd

kopiczko:


cmd/prepare/runner.go
```diff
@@ -23,11 +20,12 @@ const (
 func runPrepareRelease(cmd *cobra.Command, args []string) error {
 	var err error
 
-	absoluteCurrentFolder, err := gitrepo.TopLevel(context.Background(), ".")
-	if err != nil {
-		return microerror.Mask(err)
+	var repositoryName string
+	{
+		o := cmd.Flag("organisation").Value.String()
```
fiunchinho, line 25: We should probably validate them and fail if they are empty.

kopiczko: They have defaults set https://github.com/giantswarm/architect/blob/203b18721c72e8befa45f4c2957eb5c0144146d4/cmd/root.go#L60-L61

PR #432, opened 2020-05-13
Revert "Removing app-operator from deployment (#407)"

tomahawk28:
This reverts commit 9af8772e7a40ef7c5c75e4671faf315d5d2b7205.

Toward https://github.com/giantswarm/giantswarm/issues/10798

We need to bring `app-operator` with thiccc branch running in CPs.

### DIFF
--- a/events/events.go
+++ b/events/events.go
@@ -13,6 +13,7 @@ var (
 	// services used in all our installations
 	baseProjectList = []string{
 		"api",
+		"app-operator",
 		"cert-exporter",
 		"cluster-operator",
 		"credentiald",


kopiczko:
How many version bundles do you have in the thiccc branch?

tomahawk28:
@kopiczko It has just one versionbundle, quite not thick ironically.

kopiczko:
> @kopiczko It has just one versionbundle, quite not thick ironically.

So why not releasing it as flat?

tomahawk28:
@kopiczko I would like to keep app-operator as unique, but at the same time I would like to keep the same repo as thiccc for serving to boot a chart-operator. So far we only came up with a solution with thiccc branch. WDYT?

rossf7:
@kopiczko Once the flat app-operator pod uses helm 3 we need the thiccc pod to bootstrap chart-operator with helm 2.

Once all tenant clusters use helm 3 the thiccc pod will be removed again.

PR #440, opened 2020-05-23
project version is read from project.go if it exists

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/10039

Our current approach is to show `$latestTag-sha1`, and we talked about using `$bumpedLatestTag-sha1`. That way we are assuming than incrementing by 1 is fine. Maybe it's not. Maybe I'm creating a PR with breaking changes and I updated my `project.go` file accordingly, so I went from `1.2.3-dev` to `2.0.0-dev`. I don't want architect to tell me `1.2.3-1a2b3c...`, I want architect to tell me `2.0.0-1a2b3c...`.

Why try to figure out what's the version instead of reading that information from a file that already contains that information?

When project has no `project.go` file we can make a "best effort" and use the "closest tagged parent" approach.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Updated `giantswarm/app` dependency to 0.2.1 (first go-modules version).
+- `architect project version`: reads version from `project.go` file if it exists, uses git if it doesn't.
 
 ## [1.0.0] 2020-04-23
```
kopiczko, line 16: First of all I’m sorry for the long comment but this is a thing that I deeply care for and I don’t mean to discourage anyone.

Correction: it isn’t latestTag but tag of the nearest tagged parent. Which is a big difference.

This is wrong place to create the change if we want it. It should be created in `gitrepo` in `ResolveVersion` function and vendored here and in `opsctl`.

> Why try to figure out what's the version instead of reading that information from a file that already contains that information?

When designing this initially we wanted to reduce confusion by being compatible with go modules pseudo versions to reduce the confusion. And we failed because we thought it would be difficult to achieve. Time passed, we gained experiences and I believe this is entirely possible right now. And I also believe this is the way to go because that reduces confusion which is already there.

This change adds a new conditional branch and additional source of truth with it which makes things more complex. Besides making the whole thing more difficult to explain and understand it makes it more fragile and harder to evolve if we start to rely on those conditional branches.

So IMO we should find a solution that is either:
* Easy to comprehend/explain.
* Solves some real problem and ideally doesn’t add up to the complexity.

If there is a real problem behind that you are trying to solve (and I feel like there is), let’s talk about the problem and think how to approach it.

I won’t block that but I wouldn’t like to proceed without broader SIG releng agreement.

fiunchinho: I like the idea of mirroring what go does when trying to figure out the version, so that go pseudo-versions and `architect versions` look the same.

I think we already have two sources of truth when defining which version should a commit be part of. We already have that in `project.go`. 

It makes sense for Go tooling to try to figure out a version because they don't enforce projects to have a file defining the current version, like other dependency systems do (maven with the pom file). We have that file in some projects, though.

To me it would be better:
- Make all giantswarm projects to have a project.go file.
- Drop `project.go` from giantswarm projects. Rely only on _pseudo-versions_ generated by `architect` (passing the version to reconcile as parameter when installing the operator).

fiunchinho: > Drop project.go from giantswarm projects. Rely only on pseudo-versions generated by architect (passing the version to reconcile as parameter when installing the operator).

instead of using parameters, it could be injected at build time like we do with other values

fiunchinho:
We decided to read the version from a different file.

PR #442, opened 2020-05-25
Add prepare-release command

kopiczko:


cmd/preparerelease/internal/modifier_test.go
```diff
@@ -0,0 +1,6 @@
+package internal
+
+import "testing"
+
+func Test_modifier_addReleaseToChangelogMd(t *testing.T) {
```
kopiczko, line 5: They will come as a separate PR.

cmd/preparerelease/runner.go
```diff
@@ -0,0 +1,62 @@
...
+			return microerror.Mask(err)
+		}
+	}
+
+	err = m.AddReleaseToChangelogMd()
+	if err != nil {
+		return microerror.Mask(err)
+	} else {
+		cmd.Printf("File CHANGELOG.md updated.\n")
+	}
```
fiunchinho, line 50:
```suggestion
	err = m.AddReleaseToChangelogMd()
	if err != nil {
		return microerror.Mask(err)
	}
	
      cmd.Printf("File %#q updated.\n", changelogFile)
```

cmd/preparerelease/runner.go
```diff
@@ -0,0 +1,62 @@
...
+
+	err = m.UpdateVersionInProjectGo()
+	if internal.IsFileNotFound(err) {
+		// Fall trough. Some projects do not have project.go file.
+	} else if err != nil {
+		return microerror.Mask(err)
+	} else {
+		cmd.Printf("File project.go updated.\n")
```
fiunchinho, line 58:
```suggestion
		cmd.Printf("File %#q updated.\n", versionFile)
```

cmd/preparerelease/internal/modifier.go
```diff
@@ -0,0 +1,180 @@
...
+		newVersion: config.NewVersion,
+		repo:       config.Repo,
+		workingDir: config.WorkingDir,
+	}
+
+	return m, nil
+}
+
+func (m *Modifier) AddReleaseToChangelogMd() error {
+	file := "CHANGELOG.md"
```
fiunchinho, line 48: Could we use a shared constant for this and `project.go`?

cmd/preparerelease/internal/modifier.go
```diff
@@ -0,0 +1,180 @@
...
+
+	err := modifyFile(filepath.Join(m.workingDir, file), modifyFunc)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (m *Modifier) addReleaseToChangelogMd(content []byte) ([]byte, error) {
```
fiunchinho, line 59: Do these functions need to have a pointer as receiver? I was thinking that it could enough having `func (m Modifier) addReleaseToChangelogMd`

kopiczko: Technically it doesn't matter but we always use pointers for that so I'd rather stay consistent. Is there a reason to change that?

cmd/preparerelease/internal/modifier.go
kopiczko: Done.

cmd/preparerelease/internal/modifier.go
fiunchinho: Just to avoid mutable data structures when no required, but we can use pointers.

kopiczko: I see what you mean. But (besides doing that everywhere else in our codebases) in case we need to hold some state in the object I'd rather be prepared for that instead of relying on the fact there will be always a fresh copy in the receiver.

kopiczko:
I'm happy with how it works:

- Initial changelog link: https://github.com/kopiczko/test-gh-workflows/pull/27/files
- Subsequent changelog link: https://github.com/kopiczko/test-gh-workflows/pull/27/files
- Rendered changelog with links: https://github.com/kopiczko/test-gh-workflows/blob/431bb67ea691133f875b7533d68d5708bf170b40/CHANGELOG.md

PR #445, opened 2020-05-28
Prepare release for 1.1.0

omegas27:
Towards: https://github.com/giantswarm/giantswarm/issues/10897


kopiczko:
Having that: can we make breaking changes to architect already?

PR #448, opened 2020-06-02
Release v1.1.1

rossf7:
So we use 0.0.0 as the version for control plane app CRs.


rossf7:
Going with a ping. So app collections don't revert to using 1.0.0 for app CRs.

PR #449, opened 2020-06-03
Revert giantswarm/app to 0.2.1 and use 1.0.0 as version for app CRs

rossf7:
We need to revert so app CRs work with app-operator 1.0.3.


rossf7:
Going with a ping.

PR #455, opened 2020-06-12
Added dingo installation

whites11:
Added new dingo installation for DVAG


kopiczko:
@whites11 can be closed?

PR #457, opened 2020-06-23
Use updated app library with apiextensions change to improve app CRs

rossf7:
Towards giantswarm/giantswarm#11556


kopiczko:
@rossf7 should we merge this?

rossf7:
It wasn't working correctly when I tested. The ugly empty fields were still there. I'll take another look this week.

kopiczko:
How about this https://github.com/giantswarm/architect/pull/472 ? (I just added dependabot config)

rossf7:
> How about this #472 ? (I just added dependabot config)

It won't break. So I'd say merge the dependabot PR and I'll close this?

I want to get back to this but I don't have time to work on it at the moment.

kopiczko:
#472 merged.

PR #488, opened 2020-08-10
Handle fix releases like 0.1.0-2

corest:
So we have these kinds of releases like `0.1.0-1` which are used to replace respective release version (e.g. `0.1.0` in this case). 

1. `architect prepare-release` fails to parse flag value when you use such version.
2. if such version is used, `project.go` should not be changed as otherwise e.g. operator is not able to reconcile CRs as it uses wrong version

This PR fixes those cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed
+### Added
 
 - Handle release versions like `0.1.0-1` in `prepare-release` command.
 - Do not update version in `project.go` file for replacement releases (versions `0.1.0-x`).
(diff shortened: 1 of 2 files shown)


corest:
Will wait for feedback from @kopiczko to be sure :D

cmd/preparerelease/internal/modifier.go
```diff
@@ -172,12 +172,25 @@ func (m *Modifier) updateVersionInProjectGo(content []byte) ([]byte, error) {
...
+	//
+	// Skip version update as it must stay the same
+
+	version := regexp.MustCompile(`[0-9]+\.[0-9]+\.[0-9]+-([0-9])+`)
+	if version.MatchString(m.newVersion) {
+		return content, nil
+	}
+
+	fmt.Println("no match")
+
```
kopiczko, line 187:
```suggestion
```

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
```
kopiczko, line 10:
```suggestion
### Added
```

PR #489, opened 2020-08-11
remove camel and centaur

calvix:
remove camel - installation test is finished for now
remove centaur- it's gone for long time already


kopiczko:
I don't know if it was already done but - it needs to be now releases and collections repos circle configs should be updated.

PR #492, opened 2020-08-11
Ignore reference versions in helm template validation

corest:
Another place where we need to skip reference version validation


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Chagned
+
+- Ignore reference versions in `helm template` command.
```
kopiczko, line 12:
```suggestion
### Fixed

- Fix `helm template` rendering for reference versions in.
```

PR #495, opened 2020-08-12
Remove suffix from reference version in project.go

corest:
Towards https://github.com/giantswarm/giantswarm/pull/12732


xh3b4sd:
A general note, code like this should have unit tests. At least I personally would be more confident about the business logic itself.

corest:
Added unit tests

PR #532, opened 2020-10-20
added new installation `gremlin`

whites11:


calvix (approved):

calvix:
gremlin

events/events.go
```diff
@@ -71,6 +71,7 @@ var environmentProjects = map[Environment][]string{
 	"gollum":    azureProjectList,
 	"gorgoth":   kvmProjectList,
 	"gorilla":   awsProjectList,
+	"gremlin":   azureProjectList,
```
kopiczko, line 74: @giantswarm/chapter-sre what do you think about maintaining those lists in `*-collection` and `draughtsman` repositories? Do we need that anywhere else?

PR #551, opened 2020-12-01
Add --config-version flag to `create appcr`

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/14603

Waiting for https://github.com/giantswarm/app/pull/50

### DIFF
--- a/cmd/create/appcr/validate.go
+++ b/cmd/create/appcr/validate.go
@@ -1,0 +1,22 @@
+package appcr
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/giantswarm/microerror"
+)
+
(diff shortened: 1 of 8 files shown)


cmd/create/appcr/validate.go
```diff
@@ -0,0 +1,22 @@
+package appcr
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/giantswarm/microerror"
+)
+
+func validateConfigVersion(v string) error {
```
kopiczko, line 10: This validation is pretty aggressive but I think it's ok. We can relax it if there is a real use-case.

kopiczko:
RFR

PR #566, opened 2021-01-11
Update giantswarm/app

kopiczko:


kubasobon:
I thought we were removing this functionality in `giantswarm/app v4.2.0`. Does it make sense to update `architect` before `v4.2.0` is released?

kopiczko:
> I thought we were removing this functionality in `giantswarm/app v4.2.0`. What do we achieve updating `architect` before `v4.2.0` is released?

That triggered the whole discussion yesterday :) Closing.

PR #611, opened 2021-04-28
update changelog with Eagle

glitchcrab:


glitchcrab:
just merging straight in, changelog correction only.

PR #624, opened 2021-05-12
Add --config-ref-from-chart

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/16948

No changelog needed because the command was not released yet.

This add a new flag `--config-ref-from-chart`. It is mutually exclusive with `--config-ref`.

When the `--config-ref-from-chart` flag is specified config ref is read form Chart.yaml file in the path from the flag value. The old style version range like `3.x.x` is converted to the new style floating major tags `v3`.


cmd/create/argoapp/command.go
```diff
@@ -23,24 +28,46 @@ func NewCommand() *cobra.Command {
...
+	}
+	if flag.AppDestinationNamespace == "" {
+		errors = append(errors, "--app-destination-namespace is required")
 	}
+	if flag.ConfigRef == "" && flag.ConfigRefFromChart == "" {
+		errors = append(errors, "--config-ref or --config-ref-from-chart is required")
+	}
+	if flag.ConfigRef != "" && flag.ConfigRefFromChart != "" {
+		errors = append(errors, "--config-ref and --config-ref-from-chart are mutually exclusive")
+	}
```
kopiczko, line 66: When it comes to mutual exclusions I've hit `cmd.MarkFlag*` limits.

cmd/create/argoapp/runner.go
```diff
@@ -2,16 +2,61 @@ package argoapp
...
+	"strings"
 
 	"github.com/giantswarm/argoapp/pkg/argoapp"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/yaml"
 )
 
+var (
+	configVersionRangeRegexp = regexp.MustCompile(`\d+\.x\.x`)
```
kopiczko, line 19: I put the regexp here to panic immediately in case the sytnax is not correct.

cmd/create/argoapp/command.go
```diff
@@ -23,24 +28,46 @@ func NewCommand() *cobra.Command {
...
+
+	if len(errors) != 0 {
+		return microerror.Mask(fmt.Errorf("invalid flag(s): %s", strings.Join(errors, ", ")))
+	}
```
axbarsan, line 70: Very cool 👍🏻

kopiczko: Yeah I wanted all the errors printed at once instead of doing rounds in the CI.

cmd/create/argoapp/runner.go
```diff
@@ -2,16 +2,61 @@ package argoapp
...
+		var chartYaml struct {
+			Annotations map[string]string `json:"annotations"`
+		}
+
+		err = yaml.Unmarshal(content, &chartYaml)
+		if err != nil {
+			return microerror.Mask(fmt.Errorf("failed to parse yaml file %q: %s", path, err))
+		}
+
+		annotation := "config.giantswarm.io/version"
```
axbarsan, line 48: Maybe store this as a constant at the top of the file?

axbarsan (approved):

cmd/create/argoapp/runner.go
kopiczko: I can do that, but what's the benefit?

axbarsan: Maybe it's just a personal preference, but it's easier to find when you need to change it

kopiczko: But arguably easier to find during the review when it's closer to the actually changed code. I'll go with this. LMK if it makes you unhappy.

PR #673, opened 2021-09-09
Remove unused packages

kopiczko:


kopiczko:
@glitchcrab so you can stop updating events package :D

glitchcrab:
> Do we still have that anywhere in the docs?

I don't think so - I can't find it so I would assume I was just doing it from memory :sweat_smile:

JosephSalisbury:
> +0 −1,482

i'm in love

PR #700, opened 2022-01-30
Add update-changelog flag to prepare-release

tfussell:
Needed for https://github.com/giantswarm/devctl/pull/332

See https://github.com/giantswarm/actions-test/pull/46 for example

The new automation will automatically generate release notes based on commits since the last release. As a result, we can stop editing `CHANGELOG.md` for repositories using this method. This PR makes architect not attempt to update `CHANGELOG.md`. It defaults to false so shouldn't affect any existing users.

### DIFF
--- a/cmd/preparerelease/runner.go
+++ b/cmd/preparerelease/runner.go
@@ -24,6 +24,11 @@ func runPrepareRelease(cmd *cobra.Command, args []string) error {
 		return microerror.Maskf(executionFailedError, "--version flag can't be empty")
 	}
 
+	updateChangelog := cmd.Flag("update-changelog").Value.String()
+	if updateChangelog != "" && updateChangelog != "true" && updateChangelog != "false" {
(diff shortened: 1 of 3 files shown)


cmd/preparerelease/runner.go
```diff
@@ -24,6 +24,11 @@ func runPrepareRelease(cmd *cobra.Command, args []string) error {
 		return microerror.Maskf(executionFailedError, "--version flag can't be empty")
 	}
 
+	updateChangelog := cmd.Flag("update-changelog").Value.String()
```
MarcelMue, line 27: Why cast this to a string here? `Value` is a bool, so maybe I am missing something

tfussell: I was being lazy and copied and pasted the other flags. I replaced it with `GetBool` now.

fiunchinho (approved):

PR #920, opened 2024-01-17
Use Go version 1.21.6, use gsoci registry

marians:


marians:
This currently fails with

> ERROR: failed to solve: gsoci.azurecr.io/giantswarm/app-build-suite:1.2.2: gsoci.azurecr.io/giantswarm/app-build-suite:1.2.2: not found

@piontec What's the way to go here in such case?

marians:
Change to using the image suffixed `-circleci`, which is available in the new registry.

piontec (approved):

PR #1078, opened 2025-02-13
Bumping deps in Dockerfile

ljakimczuk:
Bump of dependencies.


marians (approved):

ljakimczuk:
I tested this by bumping `architect` version in `architect-orb` executor, in [this PR](https://github.com/giantswarm/architect-orb/pull/595), and then using the `architect-orb` in the `app-operator`, in [this PR](https://github.com/giantswarm/app-operator/pull/1355). All seems to work, but it also seems two of the tools I am bumping here are not used.

The `yamale` tool seems to not be used anywhere through the `architect`, so not sure it is still needed. Github search shows some used for it in individual project, but outside of the `architect` context. The tool seems to be added quite a time ago, so maybe it was an experiment or at the time we had a need for it. But I would opt in for leaving it part of the `architect` just in case we have some magic that relies on it somewhere, somehow.

The `kubeconform` tool seems to be used, when building Helm Charts and pushing to catalogs, but only when `architect` is the executor, what isn't the case these days, at least in case of majority of the projects, which tend to rather use ABS.

marians:
So sounds like fine to bump.

Maybe we should keep a note of these findings, to reduce duplication (with regard to ABS) and simplify (with regard to yamale) in the future.

ljakimczuk:
@marians sounds fine, I will add a note to the Dockerfile

PR #1195, opened 2025-08-15
Go 1.25

marians:


Dockerfile
```diff
@@ -69,7 +69,7 @@ RUN mkdir ~/.ssh &&\
 # See https://peps.python.org/pep-0668/
 RUN rm -f /usr/lib/python3.11/EXTERNALLY-MANAGED
 
-RUN pip install yamllint==${CT_YAMLLINT_VER} yamale==${CT_YAMALE_VER}
+RUN pip install --break-system-packages yamllint==${CT_YAMLLINT_VER} yamale==${CT_YAMALE_VER}
```
marians, line 72: This was needed to circumvent a warning that Python 3 emits in this case in general now.

PR #1216, opened 2025-09-29
Fix multi-arch image for arm64

marians:
Follow-up to https://github.com/giantswarm/architect/pull/1212

Previously the binary included in the arm64 architecture part of the image was not suited for arm.

This PR aims to build an image where the binary is actually an ARM one.

Test:

image


marians:
In https://app.circleci.com/pipelines/github/giantswarm/kubectl-gs/7701/workflows/ad053cc6-6fc0-4f5e-9fbe-4d66623619c0/jobs/12932 we have see this working (resource_class is "arm.medium").

piontec (approved):
Nice!

PR #1309, opened 2026-05-07
Tidy Dockerfile: redundant apk update, no-op xtrace, fragile chmod

QuentinBisson:
Three trivial Dockerfile cleanups, no behavior change to the image content.

### Changed

- **`apk update`** removed from the apk install line. `apk add --no-cache` already fetches a fresh index in memory and skips writing the on-disk cache, so the explicit update is redundant and leaves cache files in the resulting layer. Documented Alpine practice.
- **`SHELL ["/bin/bash", "-xc"]`** replaces the separate `RUN set -o xtrace` line. The previous form only set xtrace for a one-shot shell that exited immediately, so none of the subsequent `RUN` commands were actually being traced. Moving the flag into `SHELL` makes tracing cover every `RUN` after that line, matching the original intent of the `# Print command` comment.
- **`chmod 600 ~/.ssh/known_hosts ~/.ssh/config`** instead of `chmod 600 ~/.ssh/*`. The glob works today (both files exist) but breaks if `~/.ssh/` is ever empty. Explicit list makes the step robust.

The `--platform=$BUILDPLATFORM` builder-stage refactor (real cross-build perf win for `go install kubeconform`) is a separate, larger PR I'll open next.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Drop redundant `apk update` from the Alpine package install — `apk add --no-cache` already fetches a fresh index.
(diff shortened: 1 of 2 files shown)


piontec (approved):
LGTM

PR #1310, opened 2026-05-07
Use upstream kubeconform release tarball instead of compiling

QuentinBisson:
Stacked on #1309 (Dockerfile cleanups).

### Why

The previous `RUN go install kubeconform` ran on the target architecture in the final image, meaning under QEMU emulation when buildx multi-arch targets a non-host platform.

The first version of this PR added a `--platform=\$BUILDPLATFORM` builder stage to cross-compile from source. **Simpler**: kubeconform already ships pre-built per-arch tarballs on its [GitHub releases](https://github.com/yannh/kubeconform/releases) (`kubeconform-linux-amd64.tar.gz`, `kubeconform-linux-arm64.tar.gz`). Same pattern as helm, kubebuilder, nancy, gh-token.

### Diff

- Drop the kubeconform-builder stage entirely.
- Replace `RUN go install ...` with `curl ... | tar -xzf - kubeconform` against the upstream release URL.
- Renovate annotation tracks `yannh/kubeconform` releases as before.

Net: -25 LOC vs the builder-stage approach. No Go toolchain run during image build.


piontec (approved):
LGTM

PR #1314, opened 2026-05-11
Bake cosign and hadolint into the architect image

QuentinBisson:
Adds two static binaries to the architect runtime image:

- **cosign v3.0.3** — sigstore signing/verification tool. Used by `architect-orb` for keyless image/chart signing today (downloaded per-run); will also be used for binary signing in `architect-orb` v9.3. Baking it in removes ~10 lines of repeated install logic from each orb consumer site and saves a network round-trip per build.
- **hadolint v2.14.0** — Dockerfile linter. Enables a future `hadolint` command in `architect-orb` without per-run downloads.

Both are renovate-tracked. hadolint's upstream asset names use `Linux-x86_64` / `Linux-arm64` (not `linux-amd64` / `linux-arm64`), so the install step translates `TARGETARCH`.

Also extracts `gh-token` to a renovate-tracked `GH_TOKEN_VERSION` ARG — it was the last binary in the image whose version was hardcoded into the URL.

Independent of `architect-orb` PRs — the orb's cosign install step is guarded with `command -v cosign >/dev/null || install_cosign` (coming in orb v9.3), so it stays correct both before and after this image lands.


piontec (approved):
LGTM!

PR #1379, opened 2026-07-09
feat(prepare-release): aggregate RC changelogs on promotion

piontec:
## What

`prepare-release` now aggregates release-candidate changelogs when promoting to stable. When `--version` is a **stable** release that already has matching `## [X.Y.Z-rc.N]` sections in `CHANGELOG.md`, their content — plus the `Unreleased` delta that `prepare-release` just moved into the new `## [X.Y.Z]` section — is merged into that stable section, ordered by Keep a Changelog category, with a note prepended under `### Changed`.

## Why

A GitHub release page shows **only that version's own changelog section**. On an RC → stable promotion, `prepare-release` moves just the `Unreleased` delta into `## [X.Y.Z]` — usually empty, since the real changes were recorded under the `## [X.Y.Z-rc.N]` sections. That left the promoted stable release (the "Latest release" users see) showing a near-empty changelog.

This logic previously lived as a ~165-line shell/awk step in `giantswarm/github-workflows`' `create-release-pr.yaml`, running immediately after `architect prepare-release` and post-processing its output. It belongs in the binary that owns changelog writing.

## Behavior

- **Self-guarding:** a no-op for RC/dev targets (`gitsemver.IsValidStable` gate) and for stable releases without matching RC entries. Idempotent (skips if the note is already present).
- **Merge order:** `Unreleased` delta first, then each RC oldest → newest (RC number compared as an integer, so `rc.1 < rc.2 < rc.10`).
- **Fails loudly** rather than silently dropping content if the aggregated section would contain a non-Keep-a-Changelog `### H3` (the `Unreleased` delta is not gated by `validate-changelog`), or if the stable `## [X.Y.Z]` header is missing.
- Leaves RC sections, footer link references and the `## [Unreleased]` heading untouched.

## Testing

- `go test ./cmd/preparerelease/...` — 26 pass, including happy multi-RC (rc.1/rc.2/rc.10 ordering), idempotency, RC/no-RC no-ops, non-canonical-heading error, missing-header error, and empty-delta + single RC.
- End-to-end against a fixture repo via the built binary (promotion, RC no-op, stable-no-RC no-op, non-canonical failure).
- Cross-checked **byte-for-byte identical** to the original shell step on the same input.

## Follow-up

Once released, `giantswarm/github-workflows` bumps its `architect` version pin in `create-release-pr.yaml`; the shell step there is dropped (PR #239 closed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


uvegla:
There are a couple of changelog parser libraries. Maybe worth using those instead of generating a new parser.

Ran a couple of simple tests and seems to generally do what we want, so LGTM I guess.

uvegla (approved):

PR #1409, opened 2026-09-09
chore: ignore graphify-out/ (regenerable knowledge graph)

teemow:
`graphify-out/` is the machine-local knowledge graph that graphify rebuilds from a post-commit hook. It is regenerable and must never be committed. This adds the same ignore block that `giantswarm/muster` already carries.

No other files change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 architect
+
+# graphify knowledge graph (regenerable, rebuilt by the post-commit hook)
+graphify-out/


mproffitt (approved):
Auto approved by claude via `pr-triage`

During review the following nits were noticed but do not block acceptance
- Adds a leading blank line before the comment block, so line 2 is now empty — cosmetic only (.gitignore:2)

PR #1417, opened 2026-09-17
fix(ci): run golangci-lint on go 1.27

piontec:
Pre-emptive. The gate is green today; this stops it breaking when the toolchain moves.

## Problem

`pre_commit_go.yaml` installs golangci-lint `2.9.0`, which is built with go1.26 and refuses a module that targets a higher version:

```
can't load config: the Go language version (go1.26) used to build golangci-lint
is lower than the targeted Go version (1.27.0)
```

`architect` declares `go 1.26.0`, so nothing fails yet. It turns red on **every** pull request the moment renovate moves the directive to 1.27.

That already happened in `opsctl`: #2665 bumped the toolchain four weeks ago and every opsctl PR was red until [opsctl#2684](https://github.com/giantswarm/opsctl/pull/2684) landed today.

## Fix

Pin Go `1.27` and golangci-lint `2.13.2`, which is exactly what [the central template](https://github.com/giantswarm/github/blob/main/languages/go/pre_commit_go.yaml) carries. This only closes drift, so the next "align files" sync makes no further change here.

## The 12 findings 2.9.0 never reached

All `goconst`, all in `_test.go`:

| File | Count |
| --- | --- |
| `cmd/preparerelease/internal/changelog_test.go` | 1 |
| `cmd/preparerelease/internal/funcs_test.go` | 4 |
| `cmd/preparerelease/internal/modifier_test.go` | 3 |
| `helmtemplate/helm_test.go` | 4 |

A table test repeats a fixture across its cases on purpose — the input and the expectation have to sit next to each other as literals for a reader to check the case at all. `.golangci.yml` sets `goconst.ignore-tests: true`, the same setting [devctl uses](https://github.com/giantswarm/devctl/blob/main/.golangci.yml). It stays enforced on production code.

**No production code changes in this PR.**

## One difference from opsctl

opsctl also needed govet's `inline` analyzer disabled, because go1.27 added it and it reports `cannot inline: type parameter inference is not yet supported` on ordinary `slices.Contains` calls.

architect does not need that. I checked rather than guessed: I set `go.mod` to `go 1.27.0` temporarily, re-ran 2.13.2, and got the same 12 goconst findings and no `inline` diagnostic. `go.mod` is unchanged in this PR.

## Checks

`pre-commit run --all-files` with golangci-lint 2.13.2: all hooks pass.


mproffitt (approved):
Auto approved by claude via `pr-triage`

During review the following nits were noticed but do not block acceptance
- `goconst.ignore-tests: true` is repo-wide for that linter; a scoped `linters.exclusions.rules` entry matching `_test.go` would express the same intent without a global switch, if you later want goconst on some test helpers — `.golangci.yml:5`
- `yamllint -c .yamllint.yaml` reports a `truthy` warning and an indentation error on the workflow, both pre-existing and inherited from the central template (identical on `HEAD~1`); yamllint isn't wired into `.pre-commit-config.yaml`, so nothing enforces it — `.github/workflows/pre_commit_go.yaml:15`

